PR #12, opened 2016-03-24
Fix CLI input argument

marcelluzs:
After my installation I tried to use `scenedetect` command and the compiler rose a Not Defined error at line 236, thats because `file` is not a valid type. After this commit everything worked okay. PS. I am using Python 3.5.

### DIFF
--- a/scenedetect/cli.py
+++ b/scenedetect/cli.py
@@ -233,7 +233,7 @@ def get_cli_parser(scene_detectors_list, timecode_formats_list):
 
     parser.add_argument(
         '-i', '--input', metavar = 'VIDEO_FILE',
-        required = True, type = file,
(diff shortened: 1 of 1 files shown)


Breakthrough:
Great find, thank you @marcelluzs for the bugfix.

Also, an extra special thank you for being the first third-party contribution to PySceneDetect!  I'll add a new section to the [Bug Reports & Contributing](http://pyscenedetect.readthedocs.org/en/latest/contributing/) page soon, listing people who've contributed to PySceneDetect, as well as a link to their website and email address (if they/you would like).

I can also add a CONTRIBUTORS.md file if you would rather have that, although I'd rather add the list to the official documentation for more visibility (feel free to share your thoughts!).

PR #20, opened 2016-11-11
python 3 print fix

tpltnt:
Hi there,

I added parentheses to make PySceneDetect (more) compatible with Python 3.
Cheers,
tpltnt


Breakthrough:
Hi @tpltnt thanks for the fix.  Note that the code you modified is for the experimental GUI and still under development, so this will be moved to a separate branch in the future (until it's completed).

Thanks again for the pull request, and your interest in the project.

PR #22, opened 2016-12-19
Speed improvement / do not decode dropped frames

MarkusBauer:
Scene detection speed can be improved if `frame_skip` is used. 
By now, PySceneDetect advises OpenCV to load, decode and return every frame, even the dropped ones (`VideoCapture::read()`). If `VideoCapture::grab()` is used instead, the dropped frames are only loaded, but never decoded nor converted to a numpy representation. 
This change prevents useless work, making scene detection (much) faster (depending on the values of `frame_skip` and the video's codec). 

For more details, read the [VideoCapture documentation](http://docs.opencv.org/2.4/modules/highgui/doc/reading_and_writing_images_and_video.html#videocapture), in particular `read()`, `grab` and `retrieve`.

### DIFF
--- a/scenedetect/__init__.py
+++ b/scenedetect/__init__.py
@@ -210,7 +210,7 @@ def detect_scenes(cap, scene_list, detector_list, stats_writer = None,
     # If start_frame is set, we drop the required number of frames first.
     # (seeking doesn't work very well, if at all, with OpenCV...)
     while (frames_read < start_frame):
-        (rv, im) = cap.read()
+        rv = cap.grab()
         frames_read += 1
 
     stats_file_keys = []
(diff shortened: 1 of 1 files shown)


Breakthrough:
@MarkusBauer thanks for the code change, will definitely be looking into merging this push request for the next release of PySceneDetect.  Great idea, I've used these functions for synchronizing stereo cameras before, but never realized they could be applied here as well, in hopes of avoiding an extra frame copy.

Have you noticed any side effects after doing this?  I only ask because I'm not sure if multiple calls to `grab()` will change the state the same as `read()`, or at least I thought you should always follow a `grab()` call with a `retrieve()` even if you're not using the frame (I'm not completely sure about this, and it may only be of consequence to live video streams like USB cameras).

MarkusBauer:
I runned the program with this change on various h264 or wmv files, with `frame_skip` values from 2 to 30. I didn't notice any side effect / change except a decent speedup.

Breakthrough:
Thanks for your commit and feedback @MarkusBauer, pull request accepted.

PR #33, opened 2017-04-28
docs: README Correct to --csv-output

colelawrence:


Breakthrough:
Thanks for this, will be greatly updating the API documentation in the following v0.5 release, when the PySceneDetect Python API will have an initial stable release.

PR #37, opened 2017-06-22
Python 2.7 support: Changed FileNotFoundError to OSError

talkain:
The exception FileNotFoundError is not defined in Python 2.7.
Instead, except the parent exception 'OSError'
https://docs.python.org/3/library/exceptions.html#exception-hierarchy

Signed-off-by: Tal Kain <tal@kain.net>


Breakthrough:
Thanks for the commit @talkain, this change will be included in the next release of PySceneDetect.

PR #40, opened 2017-07-19
Skip retaining the last frame if images are not saved

elcombato:
If images at scene cuts are not saved, copying of the last frame is skipped.
I recorded savings in runtime of `detect_scenes()` with `scene_manager.save_images=False` between 6% and 11%.


Breakthrough:
Thanks for this PR, I'll definitely merge this into the next release of PySceneDetect.  I'm working on some major API changes at the moment, so will delay accepting this until then (just want to make sure I can reliably test the changes).

That being said, I don't see any changes to the API that will require any changes to this PR, so I should be able to merge it as-is.  Will let you know when I've merged and tested this, but it will definitely make the upcoming v0.5 release, so should be soon hopefully.

Thanks again for taking the time to help improve PySceneDetect.

Issue #44, opened 2017-08-30
Precise Splitting


cjbarth:
Is there any interest in a PR for using ffmpeg to do precise splitting of files? I know the current splitting only works on key-frames.

Breakthrough:
Hello @cjbarth;

I've been giving this some thought for a while, the only issue I see is the passing of the video encoding parameters properly to ffmpeg (or maybe providing some default values instead).  It would be possible to add additional command-line parameters to PySceneDetect, which could be passed to ffmpeg for each split (I assume there will need to be a call to ffmpeg for each individual scene, as opposed to mkvmerge, at least given my use of it - although this presents opportunities for parallelism).

Indeed though, if you have any suggestions or changes with respect to allowing ffmpeg in addition to mkvmerge, I would be most open to attempting this.  While my primary focus is creating a stable API for the next release, I can see this being a very useful feature in the future.

cjbarth:
I was thinking that some sensible defaults for ffmpeg would be all that is needed. Perhaps a CLI argument to take an integer for `qscale`, one to take a string for `vcodec` and one to take a string for `acodec`.

As for compatibility with the libav project, if we stick to the above mentioned CLI args, then compatibility will be no problem. We can just have python check for `ffmpeg` first, if it exists, use that, and if it doesn't, try to use `avconv`.

About threading, ffmpeg tries to thread as much as possible. Therefore, it would be wise to keep the usage of ffmpeg serial from the perspective of python and only start one instance of ffmpeg at a time.

How does that sound?

Breakthrough:
Sounds excellent to me from an implementation perspective, that sounds like a sensible approach for a first implementation.  Can the bitrate and number of passes be modified through the `vcodec` argument, or is that a separate parameter (or related to `qscale`)?  (I apologize, am not too familiar with `ffmpeg`).  As for `acodec`, that sounds like a good default value to me as well, and allowing an override would make sense too.

I agree with your comments about threading, scenes should just start to be encoded/outputted sequentially, in the order they are detected.  Likewise, I agree with your comments about the threading model, and it sounds good to me.  For now it would be best to just start batch processing the results one by one, exactly they are now with `mkvmerge` (once scene detection is finished).

Your approach sounds good to me, and looking forwards to supporting this as a feature, as well as any further input or PRs you might have.

----------------

[...]

PR #50, opened 2018-02-13
Add precision cut

martinDRD:
As per issue #44 I solved this problem adding the function split_input_video_precision in _init_.py . This function allows to do a more precisely splitting of the video using ffmpeg. Now you can use the flag -pr or --precision to use this this function instead of the mkvmerge function.
Finally I modified the README to add ffmpeg as an optional dependency.


Rapsey:
I'm interested in this feature so I decided to try it, but unfortunately it seems to be bugged for me:

- The first few seconds of every clip don't play any video
- The last frame of every clip is the first frame of the next one (i.e. clips are 1 frame too long)

I am able to create "perfect" clips manually with ffmpeg by using the timestamps from the stats file:

    ffmpeg -i input.mp4 -ss 00:08:58.760 -to 00:09:18.440 -c:v libx264 -c:a aac clip.mp4

For me this results in clips that include the entire segment from the first frame to the last, with no extra frames (using the content detection method, there is no black between the segments).

Breakthrough:
When you say the timestamps from the stats file, do they correspond with the output from PySceneDetect, or are you choosing different frames manually for those timestamps?  I'm working on the v0.5 release now which includes a major update to the overall API, which should make implementing features like this significantly more easier (#57).

Rapsey:
The timestamps I used are indeed the ones selected by PySceneDetect.

From the output:

    [PySceneDetect] Comma-separated timecode output:
    00:08:58.800,00:09:18.400

From the stats file:

    13469,00:08:58.760,1.00509548611,0.956423611111,0.886793258102,0.698860677083
    13470,00:08:58.800,10.1724424913,23.2625054253,17.7659447338,19.8628862847
    13471,00:08:58.840,0.739031032986,1.00354275174,1.25472222222,2.02159288194
    (...)
    13959,00:09:18.360,0.549601779514,0.804519314236,0.788347800926,1.01092230903
    13960,00:09:18.400,44.2590581597,33.1386176215,38.0383119936,36.7172601997
    13961,00:09:18.440,4.27869032118,3.26656032986,3.52136465567,3.01884331597

Note that in my previous comment I intentionally used a time range that was 40ms (1 frame @ 25 FPS) wider in both directions. I did this to verify the precision by doing a frame dump and confirming there is exactly 1 frame too many at the beginning and the end of the clip.

The timestamps are correct. The problem is that no transcoding is done.

PySceneDetect does this:

    ffmpeg -i input.mp4 -ss 00:08:58.800 -to 00:09:18.400 -c:v copy -c:a copy clip.mp4

This essentially slices a chunk of raw data from the H264 stream. But H264 streams are a mix of I-frames (full frames / key frames), P-frames (incremental frames relying on data from the previous frames) and B-frames (relying on both previous and forward frames). In order for H264 to play properly the first frame should be an I-frame. Since PySceneDetect in precision mode slices the stream at an arbitrary point the beginning of the clip is corrupted (incremental frames can't render properly) and the video only starts playing when the first full I-frame is encountered. It appears for this particular video another side effect is that the clip contains 1 frame too many at the end.

It's just a matter of re-encoding the H264 stream. The following command creates a frame-perfect clip using the exact same timestamps. I'm only re-encoding the video because the audio is AAC which can be split just fine without re-encoding.

    ffmpeg -i input.mp4 -ss 00:08:58.800 -to 00:09:18.400 -c:v libx264 -c:a copy clip.mp4

Obviously re-encoding is not ideal (it's lossy after all) but it's pretty much required to split H264 properly. The only way to avoid it that I can think of would be to do some funky splitting & merging to re-encode only the beginning and the end of the video. FFmpeg supports segments (i.e. splitting at an I-frame) and can concatenate H264 bitstreams without re-encoding. It should be possible to split each clip into three parts: the first few frames, the middle part (containing most of the video) and the last few frames. Then you can avoid re-encoding the vast majority of the video stream: if the middle piece was split as an intact segment you could re-encode only the intro- and outro-segments and then merge all three back together into one clip. It's probably more trouble than it's worth though.

Breakthrough:
Hi @martinDRD / @Rapsey;

This has been fixed in v0.5-beta, and will be included in the release of v0.5.  See [my comment here on #44](https://github.com/Breakthrough/PySceneDetect/issues/44#issuecomment-409713035)

Issue #64, opened 2018-07-29
Performance Improvement for ThresholdDetector


bubalazi:
Hi,

Frame averaging currently is done via
```#127@detectors.py
numpy.sum(frame[:,:,:]) / num_pixel_values
```

I've experimented a bit and found that a substantial improvement can be made by using OpenCV built-in methods for this purpose. For example:
```
cv2.mean(cv2.mean(frame))[0] / 3
```
The resulting `compute_frame_average()` method runs ~10% faster on the test video.

Example:
```
# Test with the proposed change
$ scenedetect -i goldeneye.mp4  -s output.csv -l
[PySceneDetect] Detecting scenes (threshold mode)...
[PySceneDetect] Parsing video goldeneye.mp4...
[PySceneDetect] Video Resolution / Framerate: 1280 x 544 / 23.976 FPS
Verify that the above parameters are correct (especially framerate, use --force-fps to correct if required).
[PySceneDetect] Processing complete, found 0 scenes in video.
[PySceneDetect] Processed 1980 / 1980 frames read in 7.8 secs (avg 254.5 FPS).
```
```
# Test with current method
$ scenedetect -i goldeneye.mp4  -s output.csv -l
[PySceneDetect] Detecting scenes (threshold mode)...
[PySceneDetect] Parsing video goldeneye.mp4...
[PySceneDetect] Video Resolution / Framerate: 1280 x 544 / 23.976 FPS
Verify that the above parameters are correct (especially framerate, use --force-fps to correct if required).
[PySceneDetect] Processing complete, found 0 scenes in video.
[PySceneDetect] Processed 1980 / 1980 frames read in 8.6 secs (avg 229.4 FPS).
```
Unit-Testing for performance of the proposed vs existing solution

```
print(frame.shape)
print("Proposed Solution results:")
%timeit compute_frame_average_PROPOSED(frame)
print("Existing Solution results:")
%timeit compute_frame_average_EXISTING(frame)
```
results in:
```
(1080, 1980, 3)
Proposed Solution results:
100 loops, best of 3: 4.92 ms per loop
Existing Solution results:
100 loops, best of 3: 6.32 ms per loop
```

Breakthrough:
Excellent, thank you for this submission!  These improvements will be added to the next official PySceneDetect release (v0.5), and have also been merged into the current master branch.

I will leave this issue open until I have merged your changes into my development branch (v0.5-dev).

Breakthrough:
[...]

PR #67, opened 2018-07-30
FIX TypeError exception on recent OpenCV versions

bubalazi:
OpenCV v3.4+ raises a TypeError when a non-numpy array is fed to cv2.mean

This is fixed for the explicit array feed-in on line #384 in detectors.py.

This may occur also on the locations where it is assumed that a frame
returned by a cv2 capture object is always a numpy array, however this
is a safe assumption.

NOTE: I still need to check the cv version of the machine which I submitted yesterdays code from, so as to be precise where this occurs and not.

### DIFF
--- a/scenedetect/detectors.py
+++ b/scenedetect/detectors.py
@@ -281,7 +281,8 @@ class ContentDetector(SceneDetector):
                 # Image math is faster with cv2
                 absdiff = cv2.absdiff(curr_hsv, last_hsv)[:3]
                 delta_h, delta_s, delta_v = cv2.mean(absdiff)[:3]
-                delta_hsv_avg = cv2.mean([delta_h, delta_s, delta_v])[0]
+                delta_hsv_avg = cv2.mean(
+                    numpy.array([delta_h, delta_s, delta_v]))[0]
 
                 frame_metrics[frame_num]['delta_hsv_avg'] = delta_hsv_avg
                 frame_metrics[frame_num]['delta_hue'] = delta_h


bubalazi:
Nope, this is not a cv2 versioning issue - I now remember that this was a last minute pre-commit 'improvement' I made so that cv2.mean is used in favor of numpy.mean, so it went through without dev testing

antoinebrl:
I can confirm the bug and this fix does work. Well done!
For performance, I would recommend to replace `numpy.array` with `numpy.asarray` but the improvement is really minor. Would be great if the PR can be merge as soon as possible :+1:

Breakthrough:
I have been trying to expedite the release of v0.5 which includes a significantly refactored API.  My development copy is now in a stable enough state for testing, so I've merged it with master just now.  Unfortunately those changes conflict with the ones you have presented here (likewise, your previous pull request has been overwritten).  The new version has several other unrelated performance improvements and works with Python 3.6+, which also yields a significant performance gain versus running under Python 2.7.

I have thus delayed merging this and closing #64 until *after* the release of v0.5.  I apologize for the inconvenience, PySceneDetect required a significant refactor however to support future development.

That being said, if you have the time and are willing to make the required changes to the new refactored code, I will consider merging it before the v0.5 release.  If not, no worries - I will look back at your modifications and make the same changes upstream eventually.  Scene detectors are now in a separate module (scenedetect.detectors), each in it's own file (threshold_detector.py and content_detector.py).

Thanks again @bubalazi for your efforts.  Also thank you for bringing that to my attention @antoinebrl, will definitely be looking into this as well.

Breakthrough:
Closed due to duplicate changes existing in #69

PR #89, opened 2019-01-08
When no cut return the whole file as a scene

piercus:
Return one scene (which will be the size of the video) when there is 0 cut detected


Breakthrough:
Thanks for the PR!

(Sorry for the delay in merging this.)

PR #99, opened 2019-04-10
Haroldlinke vegas scenedetect

haroldlinke:
Hi Brandon,

I have created this pull request. Adding the changes I did to integrate PyScenedetect into VEGAS Video Editor. Here are my comments from the issue for your reference:
I am working on an implementation Scenedetection solution for the VEGAS Pro video editor https://www.vegascreativesoftware.com/gb/vegas-pro/.
The user can select one or several clips in the mediapool or one or several clips on the timeline and the clips are cut into pieces based on scene changes. I am using your PySceneDetect API to analys the clips.
As the user interface is a bit different froma command line I have added some features:

    

- Configuration parameters are provided via a JSON file
- In a preview window every 50th frame is shown including a progressbar (tqdm does not work in the environment)

During the adaptation I saw some oints where I think some performance improvements couldbe possible:

in the content detection you first convert colorspace to HSV, then you split and then you calculate the mean value over all panes using numpy.
I tested the following approach, that works well:
-     convert colorspace to HSV, then use the ABSDIFF function to determine the delat image, then the MEAN function to determin the mean value per color and then sum the results up as you did it.

Is there any cause why you do not use these openCV functions?
Tests showed that in many case using RGB instead of HSV provides good results too. I therefore added a configuration parameter to select if HSV or RGB should be used
I do not see the need to save the last_frame in self.last_frame. self.last_hsv is sufficient. When self.last_hsv is None (e.g. at the start of the clip) it is sufficient to set delta_avg = 0 and skip the delta_avg calculation.

I used Visual Studio for updating and debugging the code.
It seems that the VS chnaged the code so that the comparision feature does not work anymore correctly.
I have marked the minor chnages with "VEGASPython" in the code.
I have created a new content-detector named content-detector-VEGAS

Harold


Breakthrough:
Hi @haroldlinke;

How did you upload these files?  Did you use a git client?  (It's possible that the line endings might have changed or something...)  Unfortunately due to the way you submitted this PR, because it appears as though you've modified every line of code in the entire project, it's impossible to review your changes.

Would you be able to re-upload this properly?  One way you can re-play the changes is to just diff your changes with the current version of master, generating a patch, and applying it to the git repository.

I appreciate your work, but do hope you understand that this cannot be accepted as uploaded.  Thank you!

Breakthrough:
Thanks for fixing that up @haroldlinke.  Just a few more questions:

 - How do you go about actually using this with VEGAS?
 - Besides the difference in adding RGB support, can you describe what else changed in ContentDetector_VEGAS?
 - Is there a reason to use print statements instead of the logger?

There are a few other minor things I would like to clean up before merging this with master.  Would it be possible to re-target this for a separate branch (e.g. call it `vegas`)?

Thank you!

haroldlinke:
Dear Brandon,

 

uups.

I tested yesterday how to fix the issue. But I thought that this was not visible yet and I need to do a new pull request. I am surprised that it worked already now.

 

I think I have to dig into Github a bit more …

 

1.	How do you go about actually using this with VEGAS?

To explain how the solution is used in VEGAS, I have to make a longer explanation:

 

One of the unique advantages of the VEGAS Pro Video Editor is that the editor provides a complete API to control the user interface and the editing process.

The available VEGAS API supports C#, Visual Basic and JavaScript.

 

As I prefer to program in Python and there was a request by some editors if VEGAS could support Python, I create a Python-Extension for VEGAS called VEGASPython. https://www.hlinke.de/dokuwiki/doku.php?id=en:vegas_python

 

This extension integrates a Python Commandline Editor into the VEGAS user interface. 

It is possible to execute Python commands directly or execute PythonScripts. The scripts can be directly integrated into the VEGAS user interface and used like other build-in commands.

 

The current available version is based on IRONPython.  (https://ironpython.net/). IRONPython has a very deep integration with .NET and works very well. 

But it has one big disadvantage: Python modules like OpenCV that are based on C-code are not compatible with IRONPython.

 

When I was looking for a solution for scene detection I found PySceneDetect. This was exactly what I was looking for and showed the advantages of using Python in VEGAS.

As PySceneDetect is based on OpenCV it was not possible to let it run under the current VEGASPYTHON.

 

As a first solution I found an OpenCV version for .NET – Emgu CV ( http://www.emgu.com/wiki/index.php/Main_Page ).

 

I had to do a lot of changes to your code as interfaces and parameters were not 100% compatible. But finally this worked very well.

 

Just before releasing the solution I looked at the licenses and found that Emgu CV was released under GNU GPL license v3. 

Ok. I can make my code available for GPL, but the code of VEGAS, where the component would be integrated, will definitely not be open source.

So it was impossible to use this solution.

 

So I had to look for a new solution: PythonNet (https://github.com/pythonnet/pythonnet/wiki)

 

Integrating PythonNet into VEGAS was a bit more difficult than IRONPython and debugging Python code in PythonNet is a bit more tricky, but finally it worked.

I have now integrated IRONPython and PythonNet into VEGAS and can select which one to use for each Python script.

 

The solution is not released yet. I am planning to make also the source code available on Github. But this takes a bit more time (especially adding more comments to the source code).

 

VegasSceneDetect will then be one part of this release of VEGASPython. Without VEGASPython it will not work.

 

For VEGASSceneDetect I have created a GitHub repository: https://github.com/haroldlinke/VegasSceneDetect

 

This repository is still private. I have made you a collaborateur, to allow you to access the source code in advance.

 

This repository includes the file VegasSceneDetect_PN.py

This file includes the Python code that provides the interface between VEGAS and SceneDetect.

 

2.	Besides the difference in adding RGB support, can you describe what else changed in ContentDetector_VEGAS?

The main changes in ContentDetector_VEGAS are:

*	use the ABSDIFF function to determine the delta image, then the MEAN function to determien the mean value per color and then sum the results up.


*	I do not see the need to save the last_frame in self.last_frame. self.last_hsv is sufficient. When self.last_hsv is None (e.g. at the start of the clip) it is sufficient to set delta_avg = 0 and skip the delta_avg calculation.

 

3.	Is there a reason to use print statements instead of the logger?

The print statements are only used for debugging.

As I explained above, I have a compete Python commandline integrated into VEGAS. With the print statements I see immediately the result and do not have to open the logfile.

Kind regards

Harold

 

 

 

 

 

 

 

 

 

 

Von: Brandon Castellano <notifications@github.com> 
Gesendet: Mittwoch, 17. April 2019 02:16
An: Breakthrough/PySceneDetect <PySceneDetect@noreply.github.com>
Cc: haroldlinke <harold.linke@t-online.de>; Mention <mention@noreply.github.com>
Betreff: Re: [Breakthrough/PySceneDetect] Haroldlinke vegas scenedetect (#99)

 

Thanks for fixing that up @haroldlinke <https://github.com/haroldlinke> . Just a few more questions:

*	How do you go about actually using this with VEGAS?
*	Besides the difference in adding RGB support, can you describe what else changed in ContentDetector_VEGAS?
*	Is there a reason to use print statements instead of the logger?

There are a few other minor things I would like to clean up before merging this with master. Would it be possible to re-target this for a separate branch (e.g. call it vegas)?

Thank you!

PR #107, opened 2019-06-11
[FIX] VideoManager read function failed on multiple videos

ivan23kor:
**Traceback:**
>"scenedetect/video_manager.py", line 757, in read
>    ::self._downscale_factor, ::self._downscale_factor, :]
>TypeError: 'NoneType' object is not subscriptable

**Reason:**
1. Initializing a VideoManager object with multiple video paths
2. Set the downscale factor, e.g. with `set_downscale_factor()`
3. Invoke `read()` function of the created VideoManager object, e.g. via passing it to the SceneManager `detect_scenes()` as `frame_source` argument

This happens because of the bug in video_manager.py around lines 750. If one of the captured videos is over, an empty numpy.ndarray is returned on line 745 and its further use leads to the aforementioned error.
The commit contains the bug fix in video_manager.py as well as an integration test (VideoManager + SceneManager) in tests/test_video_manager.py

### DIFF
--- a/scenedetect/video_manager.py
+++ b/scenedetect/video_manager.py
@@ -739,20 +739,18 @@ class VideoManager(object):
         if not self._started:
             raise VideoDecoderNotStarted()
 
-        read_frame = False
         if self._curr_cap is not None and self._end_of_video != True:
-            while not read_frame:
+            read_frame, self._last_frame = self._curr_cap.read()
+
+            # Switch to the next capture when the current one is over
+            if not read_frame and self._get_next_cap():
                 read_frame, self._last_frame = self._curr_cap.read()
-                # Switch to the next capture when the current one is over
-                if not read_frame:
-                    # Break the loop when all the captures are over
-                    if not self._get_next_cap():
-                        break
-                    # Get frame of the new capture
-                    read_frame, self._last_frame = self._curr_cap.read()
-                if self._downscale_factor > 1:
-                    self._last_frame = self._last_frame[
-                        ::self._downscale_factor, ::self._downscale_factor, :]
+
+            # Downscale frame if there was any
+            if read_frame and self._downscale_factor > 1:
+                self._last_frame = self._last_frame[
+                    ::self._downscale_factor, ::self._downscale_factor, :]
+
         if self._end_time is not None and self._curr_time > self._end_time:
             read_frame = False
             self._last_frame = None
--- a/scenedetect/video_manager.py
+++ b/scenedetect/video_manager.py
@@ -743,8 +743,13 @@ class VideoManager(object):
         if self._curr_cap is not None and self._end_of_video != True:
             while not read_frame:
                 read_frame, self._last_frame = self._curr_cap.read()
-                if not read_frame and not self._get_next_cap():
-                    break
+                # Switch to the next capture when the current one is over
+                if not read_frame:
+                    # Break the loop when all the captures are over
+                    if not self._get_next_cap():
+                        break
+                    # Get frame of the new capture
+                    read_frame, self._last_frame = self._curr_cap.read()
                 if self._downscale_factor > 1:
                     self._last_frame = self._last_frame[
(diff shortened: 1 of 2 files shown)


scenedetect/video_manager.py
```diff
@@ -743,8 +743,13 @@ def read(self):
         if self._curr_cap is not None and self._end_of_video != True:
             while not read_frame:
                 read_frame, self._last_frame = self._curr_cap.read()
-                if not read_frame and not self._get_next_cap():
-                    break
+                # Switch to the next capture when the current one is over
+                if not read_frame:
```
Breakthrough: I think this should be changed to `while not read_frame` instead now, since if for some reason the new cap doesn't return a frame, it will still be used below.  Changing this to a while() would advance to the next cap, if any.

tests/test_video_manager.py
```diff
@@ -281,3 +282,19 @@ def test_multiple_videos(test_video_file):
         # Will release the VideoManagers in vm_list as well.
         video_manager.release()
 
+def test_many_videos_downscale_detect_scenes(test_video_file):
+    """ Test VideoManager handling decoding frames across video boundaries. """
+
+    NUM_FRAMES = 10
```
Breakthrough: Unnecessary constant, can probably be removed?

Breakthrough:
Hi @ivan23kor;

Thank you so much for the fix, much appreciated!

I have one small favour to ask though - as part of the review, I think the if statement should be changed to a while to handle the case where a video doesn't have any frames for some reason.

Do you agree/disagree?  Just want to get your feedback on that before merging this.

Thank you!

scenedetect/video_manager.py
Breakthrough: Sorry I realized just now, a while loop is not the solution actually - have a better suggestion.

scenedetect/video_manager.py
```diff
@@ -743,8 +743,13 @@ def read(self):
...
+                    # Break the loop when all the captures are over
+                    if not self._get_next_cap():
+                        break
+                    # Get frame of the new capture
+                    read_frame, self._last_frame = self._curr_cap.read()
```
Breakthrough: Replace this line (line 752) with 'continue' instead - I think that will achieve what I was aiming for in the first comment I made.

Breakthrough:
Hi @ivan23kor;

Sorry my mistake - I think it would be better to just change line 752 to `continue` instead.  Thoughts?

Thanks!!

tests/test_video_manager.py
ivan23kor: Yeah, I'll leave that out in the next commit.

scenedetect/video_manager.py
ivan23kor: `cv::VideoCapture.read()` [returns](https://docs.opencv.org/3.4/d8/dfe/classcv_1_1VideoCapture.html#a473055e77dd7faa4d26d686226b292c1) (False, None) in the case where a video doesn't have any frames for some reason. Therefore, I propose to switch from while to if as one read attempt would suffice to get the next frame:
```suggestion
if self._curr_cap is not None and self._end_of_video != True:
    read_frame, self._last_frame = self._curr_cap.read()

    # Switch to the next capture when the current one is over
    if not read_frame and self._get_next_cap():
        read_frame, self._last_frame = self._curr_cap.read()

    # Downscale frame if there was any
    if read_frame and self._downscale_factor > 1:
        self._last_frame = self._last_frame[
            ::self._downscale_factor, ::self._downscale_factor, :]
```

ivan23kor:
The second commit is better, check it out!

scenedetect/video_manager.py
Breakthrough: Ah good point!  Very nice refactor, looks great :)

Breakthrough:
Very nice @ivan23kor!  Thank you so much!!

PR #110, opened 2019-07-29
Fix param in example usage

DLu:
[`detect_scenes`](https://github.com/Breakthrough/PySceneDetect/blob/master/scenedetect/scene_manager.py#L384) does not take the parameter `start_time`


Breakthrough:
Thanks for your submission.  I've changed the parameters to follow the example as in the [SceneManager API Reference](https://pyscenedetect-manual.readthedocs.io/en/latest/api/scene_manager.html).

Issue #123, opened 2019-10-24
Allow selection of which scenes images and video clips are made with save-images and spit-video


tonycpsu:
**Description of Problem & Solution**

I've had some situations where I have a large number of video files with a similar structure where I'd like to select (or omit) a certain number of scenes from the output, either at the beginning or end of the file.  These are generally Youtube downloads where the channel has a bunch of content at the beginning (title sequence) or end ("please like and subscribe!") that I'd rather not waste time saving and deleting.

It would be nice of `scenedetect` had a way to select these to avoid extra files being created.

**Proposed Implementation:**

My proposal is to add a `--scenes` (`-S`) option that takes a string in the form of a comma-separated list of array indexes or slices.  These indexes and slices would be used to generate a list of array indexes for the detected scene list, and only those indexes would be output when running `save-images` or `split-video`.

For example:

```scenedetect -i input.mp4 --scenes 0,-3:```

would save the first scene and the last three scenes, while

```scenedetect -i input.mp4 --scenes 5:-5```

would omit the first five and last 5 scenes.

In the case where the scene indexes are all positive (i.e. none of them depend on knowing how long the list is) my implementation of this also short-circuits the `detect-scenes` process, stopping after it knows that there are no more scenes that will be kept.  So `-S 0` runs extremely fast, breaking out of scene detection once it gets to the first cut, and just saving the images / video from the first scene.

Knowing a priori what scenes are to be kept is a bit of a unique situation, so I don't know if this is anything anyone else would care to have, but I've already implemented it and it works for me (tm), so I'd be happy to submit a PR if there's any interest.  Would just need to tidy it up and add docs.

Breakthrough:
I still think there's value in saying "give me the first X scenes" or "give me the last X scenes", or a combination thereof.  I just want to make sure the command-line syntax is correct or intuitive (I'm not sure I follow what the slice operator means in the first example?).

I like the idea of breaking early as well, but am curious as to how you did it - feel free to post a pull request if you would like and link it to this issue.

tonycpsu:
> I just want to make sure the command-line syntax is correct or intuitive

Understandable.  If you think that this will be a significant speedbump for users, I'm open to other suggestions on how to accomplish the same thing with a more intuitive syntax.

> (I'm not sure I follow what the slice operator means in the first example?).

[...]

PR #125, opened 2019-11-03
Support selection of scenes for `save-images` and `split-video`

tonycpsu:
Add a new `-r` / `--scene-range` command-line option to `scenedetect`
that allows users to specify the scene indxes they'd like to save images
and/or videos from.  This option should be a comma-separated list of one
or more integers and/or ranges.  Ranges are specified in a syntax
similar to python's list slice notation, e.g. `start:[stop:[step]]`.

Scene detection will stop once the number of scenes to be selected has
been detected.  (If any of the indexes is specified as a negative offset
from the end, then all scenes must be detected, and no such
short-circuiting is possible.)

See #123 for discussion.

### DIFF
--- a/scenedetect/cli/__init__.py
+++ b/scenedetect/cli/__init__.py
@@ -173,6 +173,19 @@ def duplicate_command(ctx, param_hint):
                              param_hint='%s command' % param_hint)
 
 
+def validate_slice_list(ctx, param, value):
+    try:
+        if value is None:
+            return []
+        return [
+            slice(*[int(i) if i is not None and len(i) else None
+                    for i in (s.split(":") + [None]*3 )[:3]])
+            if ":" in s else int(s) for s in value.split(",")
+        ]
+    except ValueError:
+        raise click.BadParameter(
+            'must be a comma-separated list of indexes or slices in start:[stop:[step]] notation'
+        )
 
 @click.group(
     chain=True, context_settings=CLICK_CONTEXT_SETTINGS)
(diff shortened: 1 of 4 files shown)


tonycpsu:
> I would greatly prefer that the split_video* functions as well as the _generate_images function were not modified by adding an extra argument for the scene indices. Instead, I would prefer that scene_list is modified to just "scenes", which is allowed to be a list of scenes (starting from 0), or a dict where the keys are the scene indices. Then, I would prefer that the scene_manager itself takes the proposed scene_indices, and returns a dict instead of a list in the case where it is specified.

I took a stab at making this change as requested, but I hit a bit of a snag, so I'm offering the latest commit as sort of a middle ground that I think keeps the code simpler.

My original intent in passing `scene_indices` separately (rather than just pre-filtering `scene_list` based on the indices) was to make sure that the numbers in the image filenames lined up with the scene numbers, with the caveat that `-r`'s indices are 0-based, while image and scene numbers are 1-based.  So, for example, if I passed `-r 1,2,4,`, I wanted the output images to be named `image-002`, `image-003`, and `image-005`, not `image-001`, `image-002`, and `image-003`.

Losing this association between the output images and scenes is a problem for my application, as I'm using the filenames in the images to refer back to the scenes.  (I realize there's no way to ensure the association for videos, since mkvmerge does its own thing with filenames, but that presents no problem for my application.)

So if I just did what I think you asked for here -- pass `scene_list` when `-r` isn't used and `dict(zip(scene_indices, scene_list))` when it is, I think the end result is far more complicated than just passing the two lists, because the functions taking the argument have to do different things depending on if `scenes` is a list or a dict.  The loops for iterating through the scenes end up totally different, because in one, the `enumerate` counter is used for scene numbers, while in the other, the dictionary key (from `scene_indices`) is used.  As soon as I got to implementing this, I realized it wasn't going to look very pretty.

My counter-proposal, in the most recent commit, instead does this:

```python
scene_list = [
    (s, i in scene_indices)
    for i, s in enumerate(scene_list)
]
```
and then just passes that to the methods that need it.  Those methods then do something like this:

```python
for i, ((start_time, end_time), selected) in enumerate(scene_list)])):
    # do stuff
```

This ended up creating a couple of changes in some other places where `scene_list` is used, but those were relatively minor.

The net result here, in my testing, does what I want in terms of filtering out the unwanted scenes while still maintaining the proper scene numbers in the filenames.

Breakthrough:
Some issues I realized going through this:
 - Needs to be rebased on v0.5.5 after merging your other change before this one
 - I see your point about losing context without modifying the signature of all associated functions to take the scene indices to export
 - I may need to create a new change based on your original commit

Sorry for the delay in getting to this, trying to clean up all PRs.  I will try to revert this back to your original implementation, but will likely have to re-address this for v0.6.  If that's the case, I'll create a new issue linked to this PR before closing it off so I can reference your original commit.

Breakthrough:
As per #123, retargeting for v1.0 as this will require changing a lot of the code to handle appropriately.  I will close this PR for now until v0.5.x is cleaned up, and decide then if this should be implemented as in this PR, or if perhaps some more significant API modifications can be made to simplify this feature.

Thank you for the implementation.

PR #128, opened 2019-11-09
Modify `--min-duration` arguments to accept a timecode instead of frame numbers

tonycpsu:
* Add new `-m`/`--min-duration` option for specifying a minimum number
  of seconds for each scene.
* Scenes shorter than this value will be merged with subsequent
  scenes, or if `-M`/`--min-duration-action` is set to `drop`, dropped
  from the scene list entirely.

### DIFF
--- a/scenedetect/cli/__init__.py
+++ b/scenedetect/cli/__init__.py
@@ -207,17 +207,6 @@ def duplicate_command(ctx, param_hint):
     ' -fs 2 processes 33% of the frames, -fs 3 processes 25%, etc...).'
     ' Reduces processing speed at expense of accuracy.')
 @click.option(
-    '--min-duration', '-m', metavar='SECONDS',
-    type=click.FLOAT, default=None, help=
-    'Ensure that scenes are at least SECONDS long, either by merging shorter scenes together'
-    ' (with `--min-duration-action=merge` or by dropping them (with `--min-duration-action=drop`')
-@click.option(
-    '--min-duration-action', '-M', metavar='ACTION',
-    type=click.Choice(['merge', 'drop']), default='merge', help=
-    'Select how short scenes are merged when using `--min-duration`.  Valid values are'
-    ' `merge` to have short scenes merged with their neighbors, or `drop` to drop short'
-    ' scenes from the output')
-@click.option(
     '--stats', '-s', metavar='CSV',
     type=click.Path(exists=False, file_okay=True, writable=True, resolve_path=False), help=
     'Path to stats file (.csv) for writing frame metrics to. If the file exists, any'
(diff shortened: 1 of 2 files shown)


Breakthrough:
I like where this PR is going in terms of maintainability, but this essentially depreciates the `--min-scene-len` option for both `detect-content` and `detect-threshold`.  I'm open to that for sure to make adding detectors easier in the future, but would like to make sure the existing detectors are modified accordingly to remove the redundant option.

(Or was there a reason you kept `-m` for the detectors as well that I might be missing?  Just want to discuss this briefly before making any decisions, thanks for the PR!)

tonycpsu:
Er, I wish I had a better answer than "I completely forgot about `--min-scene-len`, but I don't... 🙈

I do think there's merit in using seconds rather than frames, as FPS varies, and I kind of feel like most people would want to specify a number of seconds.  But that would break backwards compatibility for not much gain...

I also think the option to have short scenes prepended/appended to neighbors might be valuable to some, but it wasn't really my primary motivation for this, so maybe this was just a poorly-conceived idea.

Will check out if `--min-scene-len` works for my purposes.

Breakthrough:
One other option which might work is modifying `--min-scene-len` to accept [a timecode](https://pyscenedetect-manual.readthedocs.io/en/latest/cli/commands.html#timecode-formats) instead of number of frames.  I completely agree that there is merit in using a unit other than frames for that purpose; for transparency, the `--min-scene-len` option was added before PySceneDetect actually supported timecodes.

Now that the parsing is done for them though, modifying `--min-scene-len` to accept any of the timecode formats would be trivial (all the parsing work is done if you look at the `time` command).

Thoughts?

tonycpsu:
For now, I've just reverted my implementation and added timecode parsing for the existing option as you suggested.

I do think there's merit to allowing users to drop short scenes (it appears the current implementation just prepends them to the next scene?), and maybe some simplicity to doing it after the scene list is generated instead of in the detection code, but this'll do for now.

Breakthrough:
Thanks so much for this, @tonycpsu!  Sorry I don't have a Travis.CI setup for the project yet, so still need to run all the unit tests on this before merging, but I don't expect any issues (again thank you for submitting very small/targeted code changes, it really helps!)

Breakthrough:
Tested manually for now; still requires documentation update.  Will try to get that pushed to this PR so it can go in for this weekend.

Sorry for the delay, and thanks again for your submission!

Breakthrough:
Documentation updated in latest ( https://pyscenedetect-manual.readthedocs.io/en/latest/cli/detectors.html ).  Default value for both -m options changed from 15 frames to 0.6 seconds.  Feel free to provide any comments if you feel this new default isn't a good choice, or if you notice anything in the new documentation.

Thanks again for the pull request!

Issue #88, opened 2019-01-07
Calling Save Images from Python Interface


KevinJ90825:
Description of Problem & Solution
I would like to be able to generate images for split scenes with the Python API, in the same way you can generate them with the CLI save-images command

Proposed Implementation:
The addition of an API method in the video-splitting module, ideally, or wherever else might be deemed appropriate.

Breakthrough:
TODO: Add SaveImages command to the Python API.

Breakthrough:
Implemented global function `scenedetect.scene_manager.generate_images`, and still TODO is to add an actual `save_images` method to the `SceneManager` which will invoke the `generate_images` function.

Breakthrough:
This function will need to be modified either as part of #123 or another changeset afterwards.

Breakthrough:
After #129 is merged, work on this will resume, which also needs to write unit tests for the new global function.

Breakthrough:
This is now available in the v0.5.5 branch.  Still TODO is complete documentation and add unit tests.  It can be called via `scenedetect.scene_manager.generate_images` now.  Targeting v0.5.5 release now.

Breakthrough:
This has been included in the v0.5.5 release.  Documentation for the `save_images()` function can be found [here (link)](https://pyscenedetect.readthedocs.io/projects/Manual/en/latest/api/scene_manager.html#scenedetect.scene_manager.save_images).

PR #129, opened 2019-11-09
Add `save-images` option for ignoring a number of frames at start/end.

tonycpsu:
Add `--image-frame-margin` option, which specifies a number of frames
to ignore at the beginning and end of scenes when generating images.

This is useful if scene cuts may have frames from the previous/next
scene in them for some reason, or if there are some transition effects
or camera movement between scenes.

### DIFF
--- a/scenedetect/cli/context.py
+++ b/scenedetect/cli/context.py
@@ -204,7 +204,7 @@ class CliContext(object):
         else:
             middle_images = self.num_images - 2
             for i, (start_time, end_time) in enumerate(scene_list):
-                timecode_list[i].append(start_time)
+                timecode_list[i].append(start_time + self.image_frame_margin)
 
                 if middle_images > 0:
                     duration = (end_time.get_frames() - 1) - start_time.get_frames()
@@ -215,7 +215,7 @@ class CliContext(object):
 
                 # End FrameTimecode is always the same frame as the next scene's start_time
                 # (one frame past the end), so we need to subtract 1 here.
-                timecode_list[i].append(end_time - 1)
+                timecode_list[i].append(end_time - 1 - self.image_frame_margin)
 
         for i in timecode_list:
             for j, image_timecode in enumerate(timecode_list[i]):
@@ -641,7 +641,7 @@ class CliContext(object):
 
 
     def save_images_command(self, num_images, output, name_format, jpeg, webp, quality,
-                            png, compression):
+                            png, compression, image_frame_margin):
         # type: (int, str, str, bool, bool, int, bool, int) -> None
         """ Save Images Command: Parses all options/arguments passed to the save-images command,
         or with respect to the CLI, this function processes [save-images options] when calling:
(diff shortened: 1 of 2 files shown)


scenedetect/cli/context.py
```diff
@@ -204,7 +204,7 @@ def _generate_images(self, scene_list, video_name,
         else:
             middle_images = self.num_images - 2
             for i, (start_time, end_time) in enumerate(scene_list):
-                timecode_list[i].append(start_time)
+                timecode_list[i].append(start_time + self.image_frame_margin)
```
Breakthrough: Would prefer bound checks on the timecodes being used to make sure they do not exceed the frame boundaries.  Specifically, if the calculated start/end times would exceed the half-way point, I think that should be defined behavior - what that behavior should be, though, I'm not 100% sure on, so any feedback or opinions you have would be welcome.

Breakthrough: Also I know I'm essentially asking you to do work for free, so if you are unable to support any of the requested changes I've made on this PR or any other ones, please don't hesitate to let me know, and I'll take over the rest of the PR accordingly.

Very pleased with the quality of your commits thus far though (very small and targeted changes), sorry I don't have more time to actively support them as much as I used to.  Thank you again for your PRs and issue requests, these are definitely things I want to add in v0.5.2 if possible (hurray for non-breaking changes :))!

tonycpsu: See what you think about a393efc.   It's a pretty significant rewrite of the code that builds the timecode list for images, but I think it accomplishes what you want by dividing each scene's frames into `--num-images` chunks and then selecting the appropriate frame from each.

The `min` and `max` make sure that the frame for the first and last images never go past the frames for the middle images.  Lightly tested with some crazy high values of `--image-frame-margin` and it seems to work as expected.

Breakthrough: Awesome, have you tested with values of 0 and 1 by any chance?  Also I'd like to write some unit tests for this eventually as well, but won't let that keep this from being merged.

Would like to do some testing on this myself before merging, but at a first glance your implementation looks a lot cleaner than mine, thank you! 😄 👍

Breakthrough:
After re-reviewing the changes, this will likely be merged as-is, with unit tests being written as part of #88.  Thanks again for the PR, will get this merged by end of the week!

scenedetect/cli/context.py
```diff
@@ -190,43 +193,52 @@ def _generate_images(self, scene_list, video_name,
...
                 self.video_manager.seek(image_timecode)
                 self.video_manager.grab()
                 ret_val, frame_im = self.video_manager.retrieve()
                 if ret_val:
                     file_path = '%s.%s' % (filename_template.safe_substitute(
                         VIDEO_NAME=video_name,
                         SCENE_NUMBER=scene_num_format % (i + 1),
-                        IMAGE_NUMBER=image_num_format % (j + 1)),
+                        IMAGE_NUMBER=image_num_format % (j + 1),
+                        FRAME_NUMBER=image_timecode.get_frames()),
```
Breakthrough: Need to add this to the documentation, but I can do that before merging.

Issue #136, opened 2019-11-24
list-scenes "CSV" file isn't really a CSV file


tonycpsu:
The file produced by `list-scenes` starts with a line containing the cut list, which means it isn't really a valid CSV file.  While it's handy to have a concise list of cuts in addition to the full scene data, I think it makes more sense for the default output to just be the CSV, and add an option to instead (or in addition) write a separate file with only the scene cut timestamps.

Maybe something like a `-f` option for `list-scenes` that takes a value of `scenes`, `cuts`, or `both`?

Breakthrough:
Hi @tonycpsu;

Are you referencing [RFC 4180](https://tools.ietf.org/html/rfc4180) here?  I agree it's not really a good idea to have different numbers of columns per row in a CSV file.

Thoughts on just outputting two files by default?  Otherwise a format argument to specify scenes/cuts works as well.  Any comments/suggestions would be most welcome.

Thanks!

tonycpsu:
It's not just that the number of colums in each row that's wrong that seems off, it's that the first line isn't really the CSV header -- the second line is.  The first line is just carrying extra information that should ideally be in a separate file.  Switching to a more flexible format like YAML or JSON doesn't really make sense, since they're harder to inspect visually, and the keys would be repeated, bloating the size of the file.

I also noticed tonight that the stats file has the same thing going on with the FPS at the top.  I'm not going to be a purist here -- it's easy enough for the tools I'm writing to skip over these lines, and in the case of FPS, it's kind of handy that it's there in the stats file.  But I figured I'd ask about this while I was in there to see if some other way to go here.

My thinking is there should be 3 files here.  The scenes CSV, the stats CSV, and a third file in YAML or JSON (your choice -- I prefer YAML but JSON comes without an extra dependency.)  The third file would contain the cuts and the framerate, and could in the future contain anything else that's needed.

This isn't super high on my list right now, but it's something I could throw together if there's interest.

Breakthrough:
Thanks for your comments, I think being a purist about this kind of stuff is somewhat important to address ongoing technical debt - quite frankly I didn't give much thought to automated analysis of the output CSV files when I first implemented those features (I was still doing a lot of manual analysis using spreadsheet software).  Having the first row not being actual headers is indeed confusing, and definitely wasn't the best long-term choice.

[...]

PR #138, opened 2019-11-28
Remove extraneous rows from CSV files.

tonycpsu:
Per #136 , this removes cut list and frame rate from top of CSV files.  Pretty sure this breaks some of the test code that looks for those things (at least the frame rate), and not really sure whether to rip out the tests or rewrite them.


Breakthrough:
Now that I have a CI environment working, will need to fix the tests before merging this, but shouldn't be too much trouble.  Will definitely want to merge this in for the next release.  My apologies for the delay in getting to this!

Breakthrough:
Unfortunately this needs to be redone due to refactoring of the changed files, but it shouldn't take too long.  Will re-apply these changes in the v0.5.5 release (issue updated accordingly).

PR #141, opened 2019-11-29
Add --drop-short-scenes option, make --min-scene-len a top-level option.

tonycpsu:
* Add `-d`/`--drop-short-scenes` to allow scenes shorter than
  `min-scene-len` to be dropped instead of combining them.
* Move `-m`/`--min-scene-len` from `detect-content` and
  `detect-threshold` and make it top-level option that both commands
  respect.

### DIFF
--- a/scenedetect/cli/context.py
+++ b/scenedetect/cli/context.py
@@ -341,6 +341,13 @@ class CliContext(object):
         # files with based on the given commands (list-scenes, split-video, save-images, etc...).
         cut_list = self.scene_manager.get_cut_list(base_timecode)
         scene_list = self.scene_manager.get_scene_list(base_timecode)
+
+        if self.drop_short_scenes:
+            scene_list = [
+                s for s in scene_list
+                if ( s[1] - s[0] ) >= self.min_scene_len
+            ]
+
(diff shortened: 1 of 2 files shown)


scenedetect/cli/context.py
```diff
@@ -341,6 +341,13 @@ def process_input(self):
         # files with based on the given commands (list-scenes, split-video, save-images, etc...).
         cut_list = self.scene_manager.get_cut_list(base_timecode)
         scene_list = self.scene_manager.get_scene_list(base_timecode)
+
+        if self.drop_short_scenes:
+            scene_list = [
+                s for s in scene_list
+                if ( s[1] - s[0] ) >= self.min_scene_len
```
Breakthrough, line 348: Does this actually combine short scenes with their neighbors, or just drop them entirely?  I'm thinking maybe this PR should be split up into two, one to make `--min-scene-len` a top level option, and another which focuses on implementation of `--drop-short-scenes`... Thoughts?

Thanks for the PR, and sorry for the lengthy delay in getting to this.

Breakthrough:
Merged into branch v0.5.5.  Thank you for the PR!

PR #144, opened 2019-12-23
resize grabbed frames, pass frame rate to detectors

allenday:


requirements.txt
```diff
@@ -3,6 +3,8 @@
 #
 opencv-python
 numpy
+pandas
+scipy
```
Breakthrough: What is the purpose of these imports in this PR?  I don't really want to expand the dependencies of the package without much discussion.

Breakthrough:
Needs to be rebased against master.  Would also prefer to split into two PR's, one for resizing based on downscale factor, and another to pass framerate to postprocess.  Closing until updated accordingly.

Issue #140, opened 2019-11-28
Include link in HTML table


allenday:
- Link to video in an additional column.
- Link each `<img/>` element to open the full image.
- Do both of these `<a/>` elements wth `target=blank`

Breakthrough:
It seems like @wjs018 addressed both of your second points (many thanks!).  As for the link to a video, there should be a command-line flag to turn on/off that behaviour.  Do you still require this modification?  Thanks!

Breakthrough:
Closing until feedback is provided, just to clean up the backlog of open issues.

Issue #139, opened 2019-11-28
Image filename in HTML needs to be URI escaped


allenday:
Image filenames can contain special characters (such as `?`). These need to be URI escaped to correctly render the image in the HTML.

allenday:
https://github.com/Breakthrough/PySceneDetect/blob/ed69e21ffa5881ef63652644d864248e5872cb14/scenedetect/thirdparty/simpletable.py#L110

Breakthrough:
Closed by #148 (thanks @wjs018!).

PR #148, opened 2020-01-28
Added links and filename sanitation to html output.

wjs018:
Updated html output code to sanitize filenames (#139) and to make the images hyperlinks to the full-sized images (#140).


Breakthrough:
Thanks for the PR, @wjs018!

Issue #72, opened 2018-08-31
PySceneDetect v0.5 New CLI & API Feedback


Breakthrough:
Hello all;

Please free to leave any feedback or suggestions you might have regarding the latest release of PySceneDetect (v0.5).  This release includes a significantly refactored Python API as well as a completely redone command-line interface (CLI).

Feedback for both the new API and CLI is requested here.

Thank you all again for your help and support with respect to preparing this release - really appreciate everyone's feedback!

wjs018:
Hi @Breakthrough,

I have been doing some work recently with the 0.5 API and overall, I think the new structure is really well thought out. There is just one minor thing I want to point out. Would it be possible to bring the `StatsManager` class into the base `import scenedetect` call? As it is right now, this is what getting ready to analyze a video looks like:

```python
import scenedetect

video_mgr = scenedetect.VideoManager([video_file])
stats_mgr = scenedetect.stats_manager.StatsManager()
scene_mgr = scenedetect.SceneManager(stats_mgr)
```

This would be fixed by adding an import statement to `scenedetect.__init__.py` to import `StatsManager` along with the `VideoManager` and `SceneManager` classes.

This would be a low priority for sure (I said it was minor). Thanks for your hard work!

Breakthrough:
Hi @wjs018;

Yes absolutely, good catch.  I intended for it to be there, I must have missed adding it.

Thanks for the report, I will definitely include this as a fix in the next release.  Likewise, thanks for your interest in the project!

Breakthrough:
Note that going forwards, the capabilities of DVR-Scan will be absorbed into PySceneDetect, with the appropriate API changes being made in v0.6 going forwards.  There may be some changes to the scene detection algorithm interfaces, but I don't think any of the changes will break existing code, just add new functionality.

That being said, thank you all for your input - I've pinned this issue to gain some last minute visibility as I plan what's going to be included in the v0.6 release (the biggest being support of [SceneStats](https://github.com/Breakthrough/SceneStats)).

Breakthrough:
Random, but API changes to the detectors will be required to implement camera flash suppression.  Specifically, a lookahead buffer will be required (of size N frames).  My idea is to allow detectors to specify, during construction, the amount of lookahead they require, and will let the SceneManager know.  That way, the SceneManager can pre-buffer that many frames and provide them to each detector, rather than one frame at a time (and requiring individual detectors to buffer frames manually, as is done currently).

[...]

PR #154, opened 2020-02-28
Added StatsManager to main import.

wjs018:
As discussed in #72, this imports `StatsManager` so that it can be called with `scenedetect.StatsManager`.


Breakthrough:
Thanks for the PR, @wjs018!

Issue #130, opened 2019-11-10
Setup Travis CI / Automated Unit Test Runs


Breakthrough:
Currently there is no continuous integration setup for PySceneDetect to automatically run unit tests on all code commits/pull requests.  Unit testing is a fully manual process, requiring developers to invoke and run the tests themselves.

This issue is to track progress on setting up a Travis CI environment for PySceneDetect, to ensure quality and functionality of the codebase, and to potentially allow for continuous deliveries in the future.  This will also help with code reviews and ensuring the quality of pull requests by eliminating unnecessary manual work.

As such, I would also like to come up with a general set of guidelines for when unit tests are expected to be added or modified when making changes.  I'm most open to anyone's feedback or opinions on the matter, and will do my best to setup a process which is easy to modify and change going forwards.

Breakthrough:
The resources branch will be removed and re-added as a submodule called "resources".  It will be moved to a new Github repository called "PySceneDetect-resources".

Breakthrough:
Completed by @joshcoales in #167 (thank you so much for your help!).  Moving resources will be considered in the future if required.

PR #167, opened 2020-05-27
Travis automated testing

joshcoales:
Hey, thought it might be handy to automate those tests you've written, so they can be run on PRs and stuff. Initially this was just adding a .travis.yml, but I needed to tweak some tests a bit too.

You can see it running here: https://travis-ci.org/github/joshcoales/PySceneDetect/branches
When you merge this, you can set them up properly here: https://travis-ci.org/github/Breakthrough/PySceneDetect

Notes:
1) The readme says we support python 3.3, but travis no longer supports that, so I've not got that being tested.
2) I changed scenedetect/thirdparty/simpletable.py because it broke python 2.7 support. So now it tries to import the python3 urllib.parse.quote, but failing that, imports python2's equivalent.
3) I fixed a number of uses of empty list default argument in simpletable.py. Do not use mutable default arguments in python: https://docs.python-guide.org/writing/gotchas/
4) I've merged the test_video_file() pytest fixture into a conftest.py file, so that all the test files can use the same one.
5) I've changed the path of the test video, so that it isn't affected by the current working directory.
6) I neatened up test_load_empty_stats() to use context managers

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,16 @@
+language: python
+python:
+  - "2.7"
+  - "3.3"
+  - "3.4"
+  - "3.5"
+  - "3.6"
+  - "3.7"
+  - "3.8"
+  - "3.9"
+install:
+  - "pip install -r requirements.txt"
+  - "git fetch"
+  - "git checkout origin/resources -- tests/testvideo.mp4"
+script:
+  - python -m pytest tests/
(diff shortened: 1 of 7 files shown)


joshcoales:
Oh, I didn't realise there's a ticket for this, also: https://github.com/Breakthrough/PySceneDetect/issues/130

Breakthrough:
Thank you so much for this, this is absolutely amazing!!

joshcoales:
Glad it's helpful!

PR #196, opened 2021-01-10
Remove first row from statsfile CSV.

Breakthrough:
Removes existing header row from statsfile CSV file to properly be an actual CSV file (see #136 for discussion).

Includes backwards compatibility so existing statsfiles can still be loaded without any errors.


Breakthrough:
Re-applied in v0.5.5 branch.

Issue #153, opened 2020-02-10
Reduce false positives in ContentDetector due to camera movement


scarwire:
Hi there! First of all, thanks for all your work on this clean, fast, and extremely promising module.

The current algorithm for detecting fast cuts with `ContentDetector` is already quite reliable, but looking through the raw data to find the correct threshold value can be tiresome, and lowering the threshold can sometimes produce false positives, especially in clips where there's no cut but there is fast camera movement. Camera movement can produce a series of frames with high `content_val`, and if any of the frames go over the threshold, it's interpreted as a cut. You can try with these examples ([1](https://danielmorgan.eu/img/retour-clip1.mp4), [2](https://danielmorgan.eu/img/retour-clip2.mp4)) that I've taken from an old film I'm studying. There are real cuts that have a `content_val` of only 9 or 10 (file 2, frame 1172, 00:46), but also false positives with `content_val` greater than 10 (file 1, frame 403, 00:16).

One possible way to deal with this would be to look for fast peaks in the amount of change -- peaks that can't last for more than a frame or two. Once we've calculated the content change for all the frames in the clip, we can then calculate the ratio of each frame's `content_val` to the `content_val` of the surrounding frames. In pseudocode:

```python
metathreshold = 3
cut_list = []
while i < len(frames):
    content_val_ratio = content_val(frames[i])
        / (content_val(frames[i - 2]) 
        + content_val(frames[i - 1])
        + content_val(frames[i + 1])
        + content_val(frames[i + 2])) / 4)
    if (content_val_ratio >= metathreshold and content_val(frames[i]) >= 5):
        cut_list.append(i)
    i += 1
```

From the tests I've done, the "metathreshold" value of 3 (mark a frame as a cut if its `content-val` is more than three times the `content-val` of its neighbors) and the baseline content-val threshold of 5 (still, don't mark anything as a cut if its `content-val` is less than 5) seem like sane defaults. I think they're easier to pick out than the current threshold value, but this is open to discussion and more testing.

I've been messing around with the code and come up with a way of implementing this as a `SceneDetector` object using the model you've already created. However, the current API calculates HSV values one by one as each frame is passed to the detector, so I also added a second post-processing step in `scene_manager.py` after all those values have been stored. This step calculates the running average of change in surrounding frames and produces a list of cuts based on this. There are probably other ways of doing this, though. 

Let me know if you think this is useful. If you're interested, I'll make a pull request.

wjs018:
[...]

PR #198, opened 2021-01-12
Adaptive Threshold Content Detector

wjs018:
As discussed in #153. This is to add a new detector `adaptive-content-detector` that dynamically adjusts the threshold of what counts as a detected scene transition based on the frames before and after it. I implemented a more fully integrated version of what @scarwire had initially put together.

I tested this detector out to make sure that nothing crashed and it was working nicely with the cli interface. I have not yet tested it on a range of different videos to assess performance or accuracy. Performance-wise, it should behave the same as `detect-content` since it is doing the same hsv delta calculations. The main difference is that a stats manager is required as it needs to re-traverse the calculated hsv stats after they have been calculated to mark detected scenes.

@Breakthrough please feel free to mark this up. One thing I am not sure about going forward is that this detector makes heavy use of the `post_process` function which I believe you had wanted to move away from in the new api.


scenedetect/detectors/adaptive_content_detector.py
```diff
@@ -0,0 +1,212 @@
...
+
+
+class AdaptiveContentDetector(SceneDetector):
+    """Detects cuts using HSV changes similar to ContentDetector, but with a
+    rolling average that can help mitigate false detections in situations such
+    as camera moves.
+    """
+
+    def __init__(self, video_manager=None, adaptive_threshold=3.0, min_scene_len=15, min_delta_hsv=5.0, window_width=2):
+        super(AdaptiveContentDetector, self).__init__()
```
Breakthrough, line 52: Technically this could inherit from ContentDetector to reduce a lot of code duplication - thoughts?  I don't mind taking that one on, so not a huge deal - it can always be done after merging anyways if that's the case.

Breakthrough (commented):
This may be one of those cases where it might be better to discuss in a persistent chat kind of thing (or face to face).  Would you be open to having a discussion on Discord (or another platform) about this sometime?

scenedetect/detectors/adaptive_content_detector.py
wjs018: I agree about chatting and happy to do so (my email should be on my gihub profile, or `doctor_walter#7001` on discord). I think the idea of inheriting from `ContentDetector` is a good idea. I don't currently have a dev environment set up, so I can work on getting that set up to poke around with it.

wjs018: The latest commit added inheritance, allowing for the `process_frame` function to be completely inherited. The trick to it is to initialize the `ContentDetector` with an impossibly high threshold so that it doesn't trigger any cuts. This leaves the cut list empty until the `post_process` function analyzes the metrics and writes the cuts.

Breakthrough: Gotcha - I think the base class method could be called instead, and have the return value discarded (rather than overriding the threshold).  Thoughts?

Breakthrough:
Looks great, thank you!!  I'll give it a more in-depth review this weekend, but don't see any major issues as-is, so will probably merge it into the v0.5.6 branch very soon.

Note that after merging will probably change the command to `detect-content-adaptive` for consistency with other `detect-` commands.  I'd also like to think if this should be a separate command entirely or just an option to add to `detect-content`, but that's something we can discuss after merging.

Breakthrough:
Thanks so much for the PR, merged!!

Issue #160, opened 2020-04-21
Resize generated images


Fookiwara:
Is it possible to add an option to save-images so that we can defined the size of the generated images?
Regards,
Phuoc

Breakthrough:
Hi @Fookiwara;

Sure, that sounds like a good idea indeed.  I might need to put some more thought into this for the long term (e.g. selecting scaling method), but I can see this definitely being a nice to have.

Thanks for the idea!

PR #203, opened 2021-01-20
Image resizing

wjs018:
Implemented saved image resizing per #160. A quick summary of how it is implemented:

- Added a scaling option, `-s` or `--scale`, to the `save-images` command that simply scales the saved images by some number while preserving aspect ratio. Values <1 result in smaller images, while values >1 result in larger images. This value, if specified, is ignored if either of the next two values are also specified.
- Added a height option, `-h` or `--height`, to the `save-images` command that saves the output images at the specified number of pixels tall. If no width is specified (see next bullet), then the width is scaled to preserve the aspect ratio. If the width is specified, then both the height and width are absolute and the aspect ratio is not necessarily preserved.
- Added a width option, `-w` or `--width`, to the `save-images` command that works the same way as the height parameter above.
- For this first pass I simply used `interpolation=cv2.INTER_CUBIC` to do the resizing. From some reading, this works pretty well for scaling up, but for scaling down it might be better to do `cv2.INTER_AREA`.
- The actual resizing happens in the `scene_manager.save_images` function
- I also tried to update help strings and documentation. Let me know if there is anywhere I missed.

### DIFF
--- a/docs/reference/command-line-params.md
+++ b/docs/reference/command-line-params.md
@@ -280,6 +280,23 @@ Options:
                         size. [default: 3]  [0<=x<=9]
   -m, --frame-margin N  Number of frames to ignore at the beginning and end of
                         scenes when saving images [default: 1]
+  -s  --scale S         Optional factor by which saved images are rescaled. A 
+                        scaling factor of 1 would not result in rescaling. A 
+                        value <1 results in a smaller saved image, while a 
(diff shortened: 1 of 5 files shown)


Breakthrough (approved):
Looks good.  I might want to split up the scaling functionality into a separate function to improve readability, but it can be done after merging.  Thank you!

Issue #208, opened 2021-03-01
ffmpeg-args Example from Command Line Usage / getting started does not work


mfeldt:
On https://pyscenedetect.readthedocs.io/en/latest/examples/usage/ there is the example

`scenedetect --input my_video.mp4 detect-content list-scenes split-video --ffmpeg-args "-c:v libx264 -c:a aac"`

Which yields

`Error: no such option: --ffmpeg-args`

wjs018:
This might be a case of outdated documentation. I have never used this feature, but looking at the code, it looks like the flag should be `--override-args`. So, try using that instead of the `--ffmpeg-args` and if that works for you I can update the docs.

Breakthrough:
Indeed this is a case of outdated/incorrect documentation.  Feel free to create a PR to update them, otherwise I can tackle this when I get a moment.

Thanks for the report!

Breakthrough:
Fixed as per #210, thanks @mfeldt / @wjs018 !

PR #210, opened 2021-03-02
Updated --override-args name in docs.

wjs018:
Updated example in docs to reflect discussion in #208.


Breakthrough:
Thanks for the PR!

Issue #216, opened 2021-05-04
FFmpeg parse error when using split-video in a loop


jffaria:
Hello! First and foremost, thanks for the great tool. Just a little suggestion for a fix here!

**Bug/Issue Description:**
When the scenedetect split-video command is used inside a loop (ie processing several video files in a directory) the FFmpeg "Parse error" is shown in the output with the filenames of the remaining files to be processed (which are then ignored). This is similar to what is observed when using ffmpeg directly in a shell loop without the -nostdin flag. I noticed ffmpeg is not called with -nostdin flag in video_splitter.py

**Required Information:**
Provide the following information to assist with reporting the bug:
1. Provide a full copy of the command line options you are using:

`scenedetect -i "$file" detect-content list-scenes save-images -p split-video `

**Expected Behavior:**
No FFmpeg error messages should be outputted and loops in the shell should work as expected.

**Computing Environment:**
 - OS: Ubuntu
 - Python Version: 3.8.5
 - OpenCV Version: 

**Additional Information:**
See this related to bash loop/ffmpeg in, eg http://mywiki.wooledge.org/BashFAQ/089 or https://www.google.com/search?q=ffmpeg+loop+parse+error+nostdin&rlz=1CADSMM_enBR747BR747&oq=ffmpeg+loop+parse+error+nostdin&aqs=chrome..69i57j33i160l2.6062j0j7&sourceid=chrome&ie=UTF-8

Breakthrough:
Note that until v0.5.6 is released, users who require a workaround can use the `-a`/`--override-args` option with `split-video` to manually specify `-nostdin`.

PR #218, opened 2021-05-04
Add -nostdin flag to ffmpeg call list in video splitter

jffaria:
Fix for #216


Breakthrough:
Thank you for the PR!

PR #233, opened 2021-08-25
Update adaptive_director.py command reference

stlalpha:
Updated docs to reflect the correct command-line command to call the adaptive detector.


Breakthrough:
Thanks for the fix up!

Issue #150, opened 2020-01-30
Ability to set fade bias to BOTH -1 and 1?


Gnishimura:
Pyscenedetect has been working great for me until I ran into some video where commercial content was taken out, leaving just a black screen for minutes at a time.  Under these situations PSD will still only make one cut (under default fade bias 0.0 it will cut at the middle, leaving a chunk of black screen on the end of scene 1 and the beginning of scene 2).  

My question is: Is there any simple way to change this behavior so that BOTH the fade out to black AND the fade in to content are counted as cuts (resulting in 3 scenes, with the middle "scene" being the entire time that the screen is blank)?

I know I could process the video twice, once with fade bias at -1 and a second time with fade bias at -2 but since it's taking me ~15 minutes to process a 1 hr long video, I really don't want to have to do two runs.

Lastly, thank you so much for this incredible open source resource, it's amazing!

wjs018:
I am not sure about the command line interface, but using the python api, it is possible to use multiple detectors simultaneously. You could create multiple detectors with different threshold parameters and analyze the video in a single pass. 

If you model your script off of the example in [the docs](https://pyscenedetect.readthedocs.io/en/latest/reference/python-usage/), you would just need to define multiple detectors and then just add them with a call to `scene_manager.add_detector(your_detector_here)` for each of them.

Gnishimura:
Thanks for the tip - this will be very helpful.  I tried implementing the two detectors with different fade biases, and it does indeed capture the black screens, but there's still an issue: 

Even when I set the min_scene_len on both detectors to 60 frames (or ~2 secs), I am getting the creation of tiny scenes that last only 20 frames or so.  Does this behavior make sense?  Other than min_scene_len, I don't see a place where I can prevent the creation of these tiny scenes.  

After inspecting the source code and I tried to make this change to the ThresholdDetector() class and run only one detector, but this didn't work either.

      `
        min_frame_for_black = 60
        if self.last_frame_avg is not None:
            if self.last_fade['type'] == 'in' and self.frame_under_threshold(frame_img):
                # Just faded out of a scene, wait for next fade in.
                self.last_fade['type'] = 'out'
                self.last_fade['frame'] = frame_num            
         
            elif self.last_fade['type'] == 'out' and not self.frame_under_threshold(frame_img):
                if (frame_num - self.last_scene_cut) >= self.min_scene_len:
                # Only add the scene if min_scene_len frames have passed.
[...]

Issue #105, opened 2019-06-11
--min-scene-len does not respect first scene


charlesvestal:
**Bug/Issue Description:**
--min-scene-len returns scenes shorter than specified length at the beginning and end of a file

**Required Information:**
Provide the following information to assist with reporting the bug:
1. Provide a full copy of the command line options you are using, for example:

`scenedetect -i video.mp4 detect-threshold --min-scene-len 14000`

2. Add `-v debug -l BUG_REPORT.txt` to the beginning of the command, then re-run PySceneDetect and **attach the generated `BUG_REPORT.txt` file**.

[BUG_REPORT.txt](https://github.com/Breakthrough/PySceneDetect/files/3276702/BUG_REPORT.txt)

**Expected Behavior:**
A clear and concise description of what you *expected* to happen.

When specifying the `min-scene-len`, I would expect no scenes to be returned of below the length specified. It appears that _internal_ scenes are respected, but not the first or last segment.

For instance, if I have a 30m video with a fade to black at 1m, 15m and 29m. I would expect with a threshold of 300,000 (10 minutes), I would only have two 15m slices. Instead, I see the first fade detected, as well as the third, as the scenes in between are of sufficient length.  

Expected behavior would be that these are included as part of the adjacent scenes.

**Computing Environment:**
 - OS: osx
 - Python Version: 2.7
 - OpenCV Version: 4.1.0_2

Breakthrough:
Hi @charlesvestal;

Thanks for the submission, will definitely add this to the list of issues to address in the next version.

Much appreciated!

Breakthrough:
Hi @charlesvestal;

I think I have a fix for this issue I've committed to master just now.  Would you be willing to test the fix out to confirm that it has resolved your issue?

Thank you!

---------------------------

You can test the fix out by cloning the repo or [downloading it as an archive](https://github.com/Breakthrough/PySceneDetect/archive/master.zip), putting the video where you cloned/extracted the files, and run `scenedetect.py` directly inside of that folder, instead of using the `scenedetect` command).

pgodwin:
Having the same issue with detect-content as well.

charlesvestal:
> Having the same issue with detect-content as well.

Possibly related to the other issue, here? https://github.com/Breakthrough/PySceneDetect/issues/106

Breakthrough:
@pgodwin good catch, thank you.  Will work on that next.

[...]

PR #250, opened 2021-11-25
Fix process_frame to make the last clip satisfy the min_scene_len

Tsunehiko:
## Problem
When splitting a video, even though the `min_scene_len` was set, a video with a shorter length than that was generated.
This problem is related to #150. (I believe that this is a problem that could not be solved with #105.)

## Cause
This is because the length of the last clip is not compared to `min_scene_len`.

## Solution
Fixed to pass the length of the entire video (`total_frames`) to `process_frame()` and compare the length of the last clip to `min_scene_len`.


Tsunehiko:
Added some changes for splitting the video starting in the middle.

Breakthrough:
Thanks for the PR.  Unfortunately this is a breaking API change, so it cannot be merged as-is (v0.5.x is closed for breaking changes now).  Coincidentally I just addressed this case in the v1.0 refactor, so this is already in place for that (short scenes will be merged with their predecessors or successors):
https://github.com/Breakthrough/PySceneDetect/blob/d654c831f7bbb1f0f03b04e27f7f67e479fd84b7/scenedetect/scene_manager.py#L559

However, I think we could achieve the same thing in v0.5 but as a post-processing stage, since the actual # of frames processed may exceed the reported video length.  This could probably be done pretty quickly here:
https://github.com/Breakthrough/PySceneDetect/blob/master/scenedetect/cli/context.py#L326

Then this logic could also be done purely within the command-line and not break the API.  I will mark this as a known issue on the v0.5 API as well.

If you wish to refactor it accordingly let me know, otherwise I can submit a patch to fix it.  Thanks for the PR.

Tsunehiko:
Thank you for your kind reply. I encountered this bug while using the Python API (not using the command line).
I used `ContentDetector` with threshold set to 7 and min_scene_len set to 180, but a 0.02sec clip was generated for a 50fps video. This bug occurred in several videos, and the common feature was that only the length of the last clip was shorter than min_scene_len. As you wrote, I think it is safer to deal with it in the post-process stage. v1.0 will solve this problem, and I think the Python API support is less important and only the command line should be dealt with in the way you suggested. I'd like to get involved, as it's a rare opportunity, but I can't make time next week, so it'll be in about 10 days, if that's okay with you? If there is no problem, I will create the issue. Finally, I'm very excited about the next v1.0.

Breakthrough:
Not a problem at all, we're glad to have people like you that are willing to get involved and help out!  I know how difficult it can be to find time to work on open source projects, so no rush.  We really appreciate your help!

One other option that came to mind was, we could add a new argument to `get_scene_list()` in `SceneManager` to also set the minimum scene length, and perform this processing stage there (along with `drop_short_scenes`).  So long as there are default values provided for both of those arguments, then that will not be a breaking change.  The benefit there is that would also fix the Python API while still maintaining backwards compatibility.

I am ok with either approach, and more than willing to consider any other alternatives you might think of or want to explore.  Many thanks for your interest and help with the project!

Breakthrough:
Closing as a fix for #266 is now in the v0.6 branch which has a new `--merge-last-scene` option to enable this behaviour.

Issue #268, opened 2022-04-20
FrameTimeCode should be identic to itself when applying the resulting values to create a new FrameTimeCode


Wirg:
Hi !

Thank you for PySceneDetect.

I ran into issues where after some serialization the resulting time frame was not the expected one.
For example:
```python
frame_time_code = FrameTimecode(61, fps=14)
seconds = frame_time_code.get_seconds()
time_code = frame_time_code.get_timecode()

...

FrameTimecode(seconds, fps=14).get_frames() == 60 # not 61
FrameTimecode(time_code, fps=14).get_frames() == 60 # not 61
```

I found the corresponding code with flooring (with `int` instead of `round`) :

https://github.com/Breakthrough/PySceneDetect/blob/f6c61dda5ddb2e1bc99471c5520cebe8a9ea7f59/scenedetect/frame_timecode.py#L227-L236

https://github.com/Breakthrough/PySceneDetect/blob/f6c61dda5ddb2e1bc99471c5520cebe8a9ea7f59/scenedetect/frame_timecode.py#L307

I ran into the following code which seems to indicate this is on purpose. What is the reason?

https://github.com/Breakthrough/PySceneDetect/blob/f6c61dda5ddb2e1bc99471c5520cebe8a9ea7f59/tests/test_frame_timecode.py#L114-L116

Have a nice day !

Breakthrough:
Hey @Wirg;

Thanks for looking into this.  It looks like some of the existing test cases are failing after your change (probably just a few off-by-one errors now), so will need to see why that is before merging #269 (thanks for the PR as well).  I'm not sure if this should be taken for v0.5 or not, as this could possibly be a breaking change, but I'd be happy to consider it for v0.6.

I ended up doing a similar workaround in v0.6 for the PyAV backend, so I think your change is correct.  Truthfully it's been quite a few years since I've touched FrameTimecode, so I can't really remember the intent behind these test cases you pointed out:

https://github.com/Breakthrough/PySceneDetect/blob/f6c61dda5ddb2e1bc99471c5520cebe8a9ea7f59/tests/test_frame_timecode.py#L114-L116

I may mark this as a known issue for v0.5, but bring your change in [for v0.6](https://github.com/Breakthrough/PySceneDetect/tree/v0.6).  Thanks for the report!

Breakthrough:
[...]

PR #269, opened 2022-04-20
fix: FrameTimeCode approximation flooring instead of rounding

Wirg:
Suggested solution to https://github.com/Breakthrough/PySceneDetect/issues/268 .

### DIFF
--- a/scenedetect/frame_timecode.py
+++ b/scenedetect/frame_timecode.py
@@ -233,7 +233,7 @@ class FrameTimecode(object):
             Integer number of frames the passed number of seconds represents using
             the current FrameTimecode's framerate property.
         """
-        return int(seconds * self.framerate)
+        return round(seconds * self.framerate)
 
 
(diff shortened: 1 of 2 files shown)


Breakthrough:
As discussed in #268, please re-target this change for the v0.6 branch, and update the failing unit tests so that the build passes.  Thanks for your help!

Breakthrough:
Also I think there are some edge cases your change might not have covered where there is still truncation happening - I'd be happy to fix that up if you aren't able to.  Thanks!

Breakthrough:
Sorry for the comment spam - I've merged your fix into v0.6 as per c5ac42b566a4c6d2c0ce64242a0cdf3b59bfe2c9.

Unfortunately I'll need to close this PR as it constitutes a breaking change for v0.5 which may cause unintended side effects.  However, this will definitely be fixed for v0.6 and will be included in the next/final beta release.

Thanks so much for the PR and suggested solution.

Issue #235, opened 2021-08-28
Split video based on CSV file


mrahmadt:
Hi 

I'm using PySceneDetect command line to generate a list of scenes, so far the result is good but some of my scenes are detected as multiple scenes.

I was thinking, is there a way to just modify the csv file generated from list-scenes (edit start/end of the frame or the timecode) and feed the csv file back again to the split-video function?

PS: I know about option -s, but it will not work in my case, some time I want to merge 2-3 actual scenes into one video during the split.

wjs018:
Are you manually modifying the csv file and then wanting to feed that csv back to the program to split the video? As the software is now, I don't think that is possible via command line, you would have to use the python api. You would have to feed the start and end timecodes for each scene to either the `split_video_mkvmerge` ([ref](https://github.com/Breakthrough/PySceneDetect/blob/148d5b4111d52f2e8fedae75ec661351d593b01f/scenedetect/video_splitter.py#L136)) or `split_video_ffmpeg`  ([ref](https://github.com/Breakthrough/PySceneDetect/blob/148d5b4111d52f2e8fedae75ec661351d593b01f/scenedetect/video_splitter.py#L202)) function depending on your environment. These python functions are really just wrappers around command line functions using either `mkvmerge` or `ffmpeg`. So, a bit of bash scripting could bridge the gap without having to dip into the python api.

An option if you want to stick with python would be `moviepy` by importing the video file and specifying a subclip using the start and end timecodes [docs](https://zulko.github.io/moviepy/ref/VideoClip/VideoClip.html#moviepy.video.io.VideoFileClip.VideoFileClip.subclip). Then, save the subclip to file [docs](https://github.com/Breakthrough/PySceneDetect/blob/148d5b4111d52f2e8fedae75ec661351d593b01f/scenedetect/video_splitter.py#L202).

This does seem like a decent feature idea though...basically feed it a scene list that you might have modified using some external means and then let it do the splitting based on that csv. Tagging @Breakthrough to see if this would be in scope or not.

Breakthrough:
I think this would be feasible enough - an option could be added to `split-video` pointing it to a scene .csv file that contains the required splits.  For now though as you said @wjs018 it might be better to just hack something together in Python that loads the scenes from the CSV and calls `split_video_ffmpeg` directly, just since there's so much in the backlog right now...  However this might be a good first issue for a new contributor to tackle.

Breakthrough:
[...]

PR #290, opened 2022-10-03
Perceptual Hash Detector

wjs018:
Hi @Breakthrough,

This is a proof of concept for a perceptual hash based detection algorithm I had started working on a while back. I updated things to be compatible with the 0.6.x updates (I think) and implemented things in OpenCV without adding additional dependencies (previously I had PIL and scipy). I also added a cli command and help strings, but have not written any actual documentation yet.

As for the actual algorithm, I re-implemented the [phash](https://github.com/JohannesBuchner/imagehash/blob/71db1d3324ec8344c507f5a89a529ff0c0156678/imagehash/__init__.py#L260) algorithm from the [imagehash](https://github.com/JohannesBuchner/imagehash) library. For a more human readable version of the algorithm, you can see the blog post [here](https://www.hackerfactor.com/blog/index.php?/archives/432-Looks-Like-It.html).

Performance wise, I chose defaults that worked well on the goldeneye clip and seemed fairly performant.

There might be some formatting issues to make it match the updated codebase (typing for instance). Feel free to change things to make it fit or, point it out and I will try to take a crack at it.


scenedetect/cli/__init__.py
```diff
@@ -613,6 +613,63 @@ def detect_threshold_command(
...
+)
+@click.option(
+    '--size', '-s', metavar='VAL',
+    type=click.IntRange(min=2), default=16, show_default=True, help=
+    'Size of the hash used in the perceptual hasing algorithm. Must be an '
+    'integer >=2.'
+)
+@click.option(
+    '--freq_factor', '-f', metavar='VAL',
+    type=click.IntRange(min=1), default=2, show_default=True, help=
```
Breakthrough, line 631: Here and elsewhere, default values must be `None` if they correspond to config file values.  This was the only way I could find to figure out if the config value or the CLI value should be used.

Instead, you can append `USER_CONFIG.get_help_string('detect-hash', 'freq-factor'))` in the help text so it still shows the default.

scenedetect/detectors/hash_detector.py
```diff
@@ -0,0 +1,168 @@
...
+            self.last_scene_cut = frame_num
+        
+        # We can only start detecting once we have a frame to compare with.
+        if self.last_frame is not None:
+            # We obtain the change in hash value between subsequent frames as 
+            # well as the actual hash value. This is refered to in a statsfile
+            # as their respective metric keys.
+            if (self.stats_manager is not None and
+                    self.stats_manager.metrics_exist(frame_num, metric_keys)):
+                hash_dist = self.stats_manager.get_metrics(frame_num, metric_keys)
```
Breakthrough, line 108: I think we should remove this optimization since long term plan is to get rid of loading stats in lieu of #235.  This is a good example of why we should remove it actually - since the metric key doesn't contain all of the parameters, if a statsfile is specified, changing them won't make the output any different.

scenedetect/detectors/hash_detector.py
```diff
@@ -0,0 +1,168 @@
...
+                curr_med = numpy.median(curr_dct_low_freq)
+                # Transform the low frequency information into a binary image based on > or < median
+                curr_hash = curr_dct_low_freq > curr_med
+                
+                last_hash = self.last_hash
+                
+                if last_hash.size == 0:
+                    # Calculate hash as above
+                    last_gray = cv2.cvtColor(self.last_frame, cv2.COLOR_BGR2GRAY)
+                    last_resized = cv2.resize(last_gray, (imsize, imsize), interpolation=cv2.INTER_AREA)
```
Breakthrough, line 133: Should refactor these calculations into a static function called "calculate_frame_hash" or something, just to reduce code duplication w/ above.

scenedetect/detectors/hash_detector.py
```diff
@@ -0,0 +1,168 @@
...
+        # Initialize last scene cut point at the beginning of the frames of interest.
+        if self.last_scene_cut is None:
+            self.last_scene_cut = frame_num
+        
+        # We can only start detecting once we have a frame to compare with.
+        if self.last_frame is not None:
+            # We obtain the change in hash value between subsequent frames as 
+            # well as the actual hash value. This is refered to in a statsfile
+            # as their respective metric keys.
+            if (self.stats_manager is not None and
```
Breakthrough, line 106: I think we can get rid of loading stats in new detectors now, since long term plan is #235 to help the majority of cases.  Using the statsfile as a cache keeps leading to subtle bugs as well if we don't keep an eye on the metric keys.  In this case, changing the hash params wouldn't work with a statsfile since we would never re-calculate the values.

That happened to me enough I figured we should strongly consider a better alternative for using the statsfile for caching calculations and just remove it for the time being to simplify the codebase.

scenedetect/cli/__init__.py
Breakthrough, line 631: Should re-use the value from `CONFIG_MAP` for `min` here and elsewhere.

Breakthrough, line 631: Default values need to be `None` in order to work with the config file (here and elsewhere).  To get the default value to show up, you can append `USER_CONFIG.get_help_string('detect-hash', 'freq-factor'))` to the `help` string.  Sorry I know this is a very non-obvious thing (gotta love business logic).

scenedetect/cli/context.py
```diff
@@ -396,6 +396,46 @@ def handle_detect_threshold(
...
+
+        if self.drop_short_scenes:
+            min_scene_len = 0
+        else:
+            if min_scene_len is None:
+                if self.config.is_default("detect-hash", "min-scene-len"):
+                    min_scene_len = self.min_scene_len.frame_num
+                else:
+                    min_scene_len = self.config.get_value("detect-hash", "min-scene-len")
+            min_scene_len = parse_timecode(min_scene_len, self.video_stream.frame_rate).frame_num
```
Breakthrough, line 420: Probably time to refactor this into a helper function (e.g. `get_min_scene_len("detect-hash")` or something) and use that here and in the other detectors.

scenedetect/detectors/hash_detector.py
```diff
@@ -0,0 +1,168 @@
...
+            
+        # If we have the next frame computed, don't copy the current frame
+        # into last_frame since we won't use it on the next call anyways.
+        if (self.stats_manager is not None and
+                self.stats_manager.metrics_exist(frame_num+1, metric_keys)):
+            self.last_frame = _unused
+        else:
+            self.last_frame = frame_img.copy()
+
+        return cut_list
```
Breakthrough, line 168: Missing newline at end of file.  Sorry the linter isn't enabled on PRs yet, I'll make a note to fix that to automate this.

Breakthrough (changes requested):
This is awesome! We need to add some tests before merging this.  There's also documentation that needs to be updated, but that can be done after in a separate change - I can help with that part too.

Before merging this, we need to:
 - [ ] Retarget for v0.6.1 branch
 - [ ] Add unit tests

--------------------

I'm curious, do you know how well this handles motion?  I saw you had used motion correction for the edge detector, so was wondering if that would be applicable here or not as well.  This PR also makes me rethink a lot of how to combine various detection methods, since now we have the possibility to incorporate all of HSV deltas, edges, and perceptual hashes.  In v0.6.1 I may move the edge detection part back to a separate detector as per your original design, rather than combining it into ContentDetector.  We could then use how confident each detector was to improve the final result.

scenedetect/detectors/hash_detector.py
wjs018: Done with 0f54efc.

scenedetect/detectors/hash_detector.py
wjs018: Done with 0f54efc.

scenedetect/detectors/hash_detector.py
wjs018: Done with 0f54efc.

scenedetect/detectors/hash_detector.py
wjs018: Done with 0f54efc.

scenedetect/cli/__init__.py
wjs018: I think I addressed this in efa4023 (I based this off what you did in `detect-threshold`, but would love you to confirm.

wjs018: I think I addressed this in efa4023 (I based this off what you did in `detect-threshold`, but would love you to confirm.

scenedetect/cli/__init__.py
```diff
@@ -613,6 +613,63 @@ def detect_threshold_command(
...
+@click.command('detect-hash')
+@click.option(
+    '--threshold', '-t', metavar='VAL',
+    type=click.FLOAT, default=100.0, show_default=True, help=
+    'Threshold value (float) that the hash_dist metric must exceed to trigger'
+    ' a new scene. Refers to frame metric hash_dist in the stats file.'
+)
+@click.option(
+    '--size', '-s', metavar='VAL',
+    type=click.IntRange(min=2), default=16, show_default=True, help=
```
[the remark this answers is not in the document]
wjs018: I think I addressed this in efa4023 (I based this off what you did in `detect-threshold`, but would love you to confirm.

scenedetect/cli/__init__.py
wjs018: I think I addressed this in efa4023 (I based this off what you did in `detect-threshold`, but would love you to confirm.

scenedetect/cli/context.py
wjs018: I implemented a version of this in 688d67e as a private function of the `CliContext` class (`_get_min_scene_len(command)`). I tested it with all the detectors (and found a typo from an earlier commit of mine in the process) and it seems to work for all of them. Let me know if you think this should be done another way.

Breakthrough (approved):
Well done, this is amazing.  Great work!

Breakthrough:
One thing I think might need to be updated is the API docs and the `scenedetect.cfg` template file.  I can follow up with those changes though, but just want to write it down so it's not forgotten.

Breakthrough (approved):

Breakthrough:
Finally got this into develop, thanks for your patience! There was some refactoring done to the CLI in the latest version, so it doesn't show up on the CLI yet.  Will need to fix that in develop branch, but I'll make sure to do that soon.

Thanks a ton for this, amazing work!

Issue #231, opened 2021-08-15
Feature Request: Adding template variables for timecodes for split-video output


LeVolve:
I asked myself if it's possible to add variables like the beginning and ending timestamp of the current clip besides of the current scene number. To save it in the filename in order to sort and organize the scenes better.
A variable like the current scene number is already available and I think it shouldn't be that diffcult. I mean in the Excel table are already these variables saved and they can use it more through that feature.

>>For example:
I have a video file named "video" and use the command
"scenedetect --input video.mp4 detect-content split-video"
The new command could looks like this
"scenedetect --input video.mp4 --output .../Folder/%FILENAME%-%CURRENT SCENE%-%START TIMESTAMP%-%END TIMESTAMP% detect-content split-video"

My output looks like this:
video-Scene-001.mp4
video-Scene-002.mp4
video-Scene-003.mp4

And it could be like this:
video-Scene-001-00;00;04-00;00;09.mp4
video-Scene-002-00;00;09-00;00;15.mp4
video-Scene-003-00;00;15-00;00;20.mp4

From this: [FILENAME]-[CURRENT SCENE]
To this: [FILENAME]-[CURRENT SCENE]-[START TIMESTAMP]-[END TIMESTAMP]
It would be very useful for sorting and searching if I can use variables in my command. Additionally it also can be added other variabe like start and end frame and many others.

Breakthrough:
This is a great idea, @LeVolve, thanks for the submission.  This is certainly possible!

The only thing I'm not sure about is the best way to approach formatting and how the user can specify that, although another CLI argument can be added to deal with that if required (my initial thought is to use :/. as separators instead of ;/, but I know the latter is more widely used in industry).

Currently there is only `$VIDEO_NAME` and `$SCENE_NUMBER`, so will look into adding the following:

 - `$START_TIME`
 - `$END_TIME`
 - `$START_FRAME`
 - `$END_FRAME`

wjs018:
@LeVolve Wanted to let you know that this has been implemented and merged into the v0.6.1 branch. Feel free to try it out. For details on the new naming options see #291. One thing to note is that this only works with the `ffmpeg` splitter backend (this is the default anyway) because `mkvmerge` does not allow the same kind of renaming of output files.

Issue #239, opened 2021-09-04
Accept braced templates instead of dollar-sign prefixed


Breakthrough:
Right now using templates in bash requires escaping the dollar signs.  Use a style similar to Python instead to help with this, e.g.:

    scenedetect [...] split-video -f {VIDEO_NAME}-Scene-{SCENE_NUMBER}.mp4

Breakthrough:
Supposedly braces need to be matched in some bash terminals as well, so might need to just use another delimiter symbol.

Breakthrough:
As found in #291, there also might be issues with only using a prefix, so using a suffix should be considered.

wjs018:
I was doing some reading on the python `Template` class and found out why the behavior identified in #291 occurs. See [here](https://stackabuse.com/formatting-strings-with-the-python-template-class/) for an explanation (tl;dr is that `_` is a reserved python character and can break substitution unless braces are also used). For example `'$VIDEO_NAME_test.csv'` does not work because the variable is followed by `_`. However, `'${VIDEO_NAME}_test.csv'` does work because the brackets separate the template variable from the `_`.

If changing the delimiter is desired, then creating a custom template class that inherits from `Template` and then rewriting the delimiter might be a good way to go. `Template` doesn't seem to support a prefix/suffix substitution scheme that I can tell, so this kind of issue with a template variable being followed by a reserved python character would require a different solution (regex substitution?).

Alternatively, this edge case could be documented in the docs with the solution of including braces.

PR #291, opened 2022-10-05
Implemented additional template variables for ffmpeg video splitter

wjs018:
Implemented the suggested template variables per #231. These new template variables are only possible for ffmpeg as mkvmerge does not support changing file output names in the same way and it is not planned for the future either (per [this issue](https://gitlab.com/mbunkus/mkvtoolnix/-/issues/1978)).

One quirk I found while testing this out is that a template variable cannot have an underscore after it. For example, using `split-video -f '$VIDEO_NAME_$SCENE_NUMBER'` will only replace the `$SCENE_NUMBER` variable and will not substitute the `$VIDEO_NAME` variable. However, if you run `split-video -f '$VIDEO_NAME-$SCENE_NUMBER'` then they both substitute just fine. I am not sure if this is a bug with python's `safe_substitute` function, but thought it was worth mentioning.

One final issue I ran into is that if the output filename has a `:` (colon) character in it, then ffmpeg crashes and will not proceed ([relevant stack exchange](https://unix.stackexchange.com/a/450305)). So, for the timecodes in filenames I replace the colons with semicolons.

I have not touched documentation for this yet, but can do so if this is to be merged.

For completeness sake, here are the new variables that are implemented:

- `$START_TIME`
- `$END_TIME`
- `$START_FRAME`
- `$END_FRAME`

The discussion in #231 also mentioned `$FRAME_NUMBER` for the `save-images` command, but it seems that one has been done already (back in [February](https://github.com/Breakthrough/PySceneDetect/commit/ca612cc68ece5205b7a1f05347dd91b6b951096a))


Breakthrough (changes requested):
Looks good to me.  Here's a quick checklist I think we should use going forwards:

 - [ ] Target for next release branch (v0.6.1 in this case, please rebase the PR)
 - [X] Ensure code has unit tests (I think this is clear enough to not need them or can be added in future)
 - [ ] Update the documentation if required

I was hoping anyone could retarget it but I think the author has to.  I'll see if there's any kind of Git hooks that can automate this.  I'm hoping to keep main/master always at the latest stable release is why, but in the long term maybe we should instead use a branch named "stable" for that purpose.

I also noticed a few of the workflows aren't running on PRs (Windows builds + code formatting checks) so I have some TODOs there as well.

Edit: I also want to see if you're able to merge this once it's approved now that you should have the correct permissions.

Breakthrough:
Forgot to mention, re:

> One quirk I found while testing this out is that a template variable cannot have an underscore after it. For example, using split-video -f '$VIDEO_NAME_$SCENE_NUMBER' will only replace the $SCENE_NUMBER variable and will not substitute the $VIDEO_NAME variable. However, if you run split-video -f '$VIDEO_NAME-$SCENE_NUMBER' then they both substitute just fine. I am not sure if this is a bug with python's safe_substitute function, but thought it was worth mentioning.

Nice find, that's very interesting indeed.  I wonder if that could be addressed in the future with #239 if we use a style that includes a suffix as well as a prefix.

wjs018:
Merged v0.6.1 branch from upstream and rebased the PR to compare against the v0.6.1 branch.

Still remaining is to update docs. Happy to take care of it in the next couple days.

Breakthrough (approved):
Approved once lint issues fixed.

camjac251:
Would it be possible to add this to the custom ffmpeg settings `-a` option as well? I have been trying to figure out how to write the starting timecode for files with ffmpeg without `-timecode` after the encoder but haven't been able to. It would be perfect if it could grab the `-ss` and set that as the starting timecode but I don't think ffmpeg has such a function, even with `-copytb 0`

Breakthrough:
@camjac251 absolutely, that's a very good idea.  Can you file a feature request for that so we can track the work?  Thanks.

Issue #243, opened 2021-09-18
Scenes.csv file should not include "Timecode List"


eksperimental:
I think having it breaks the purpose tabulated data, as it has an irregular number of columns.
I think I additional file named "*.TimecodeList" containing either one line comma separated, or one timecode per line.

eksperimental:
Additionally, the Timecode List:  string is missing the initial entry with timecode: 00:00:00.000
This is how it is shown now,
`Timecode List:,00:00:03.440,`...

Breakthrough:
This was discussed [here in #136](https://github.com/Breakthrough/PySceneDetect/issues/136#issuecomment-761883299) - you can call `list-scenes -s` to skip the first row.

This row was intended for mkvmerge, which accepts the timecodes where a new video should be created (see docs for `--split` at https://mkvtoolnix.download/doc/mkvmerge.html).

Issue #115, opened 2019-08-31
Allow for clips to be split with content outside of start/end timeframe


charlesvestal:
**Description of Problem & Solution**
When using -start and -end, the tool defines the area of a video to be processed. In some cases, however, you may wish to limit the area to be analyzed, but treat the area outside of that window as part of the split files.

Example, 30 minute file with one scene break at 15 minutes:

`0-------------13-15-17--------------30`
`--------------***[]***----------------`

In the current tool, a command such as `scenedetect -i somevideo.mp4 time -s 00:13:00 -e 00:15:00 detect-threshold split-video -c` will produce two clips, each of 2 minute duration. 

However, in some cases you may wish to have two 15 minute clips, including the content before and after the processed duration.

**Proposed Implementation:**
A proposed command line to enable this would be `-expand-split-clips` 

**Alternative Solutions:**

Using the timecodes generated can be fed into ffmpeg and executed manually / via a script.

Breakthrough:
This should definitely be possible to add to PySceneDetect, thanks for the suggestion @charlesvestal.

Technical notes:

 - It should be possible to just modify the arguments passed to the `get_scenes_from_cuts()` function called by the SceneManager, passing 0 for the start time and the video length for the end time
 - When creating the VideoManager with this option, only the start time should be modified; the end time must only be set when calling `detect_scenes()`
 - The only thing missing which makes this a relatively simple fix is a reliable way to get the total number of frames in a video file; I've found using the OpenCV method can be somewhat unreliable (thus internally it is only used for estimation of progress, but video streams are processed until exhausted)
 - Thus either something like `SeekToEnd()` needs to be implemented to the VideoManager to manually seek to the end and get this number in the case where `-expand-split-clips` is passed
 - Perhaps an additional option could be added to allow using the estimated total length instead of seeking through the whole video (which is currently somewhat slow), but I would prefer correctness over speed for the first implementation of such a feature regardless

PR #294, opened 2022-10-25
Split video based on csv file

wjs018:
As discussed in #235, #275, and #115 this is a frequently requested feature. I have put together a first pass at implementation based on @Breakthrough's outline in #275. I would not say this is in a complete state yet, but wanted to get some eyes on it and feedback as to how some of the details of this command should work. I have outlined the way it currently functions below:

The new detector (`load-scenes` command from the cli) is designed to use a csv file as input and detect scene transitions based on the scenes in the csv. It doesn't actually do any video or image analysis and it ignores things like `min_scene_len` or the use of a `StatsManager`. It also isn't really meant to be used in conjunction with other detectors (though I haven't forbidden it yet). It's main use case is for manually editing a previously generated csv file to tweak the detected scenes and then using that to split the video based on the scenes in the csv file.

```
    Command: 
        load-scenes
    
    Arguments:
        --input, -i
            Specify path to input csv file. Currently required as no default is set.
        
        --start-col, -s
            Specify the header of the column used to mark the start of new scenes. Column
            contents can be anything parseable by `FrameTimecode`. Defaults to "Start Frame".
        
        --end-col, -e
            Specify the header of the column used to mark the end of scenes. Defaults
            to "End Frame".
        
        --framerate, -f
            Framerate of the video. This is needed only if scene starts and ends
            are specified in timecodes rather than frame numbers to do conversions.
```

The way this command currently works is that it reads the input csv file row by row and as the video is processed, when the current frame number is equal to the frame number of the next scene start (per the csv), then it "detects" a cut. This means that the csv must be in order from earliest scene to latest. Also, if the start time of the video is later than the first scene in the csv, it will not detect anything. @Breakthrough, is this desired behavior? If not, then I might need some help brainstorming how to change this.

Another important bit to mention is that this detector currently can't skip over unwanted scenes. This is something that was brought up in #275, but I am not sure how to (with the current api) mark scenes to be ignored. The detectors just produce a list of cuts. @Breakthrough I am guessing this is maybe an improvement for a future api change. EDIT: Just stumbled across #123 which is a relevant discussion for this point.

As a corollary to the above point, even if the input csv file doesn't have a scene starting at frame 1, this detector (like all the others), will include a scene the begins at the first frame and goes until the first detected cut (from the csv).

One final point is that currently the `--end-col`/`-e` argument is not really used for anything. I debated just getting rid of it entirely, but some of the uses thought up in #275 would require it even though the current api (I don't think) supports them.

This command is compatible with the csv file generated by the `list-scenes` command as long as the `-s` flag is included (`list-scenes -s`). Otherwise, the column headers are not read correctly and the csv file in general is not formatted in a way to programmatically parse. See #243 and #136 for discussion on this.

If it isn't obvious by now, this isn't in a final form yet, but I wanted to get some feedback. I haven't updated any docs or tests yet but can do so if this starts to gel into something approaching a final form.


toomanynights:
Hey!
Just wanted to say that this feature is still very much required, at least in my use case.
I experiment a lot with this amazing tool, but each time something goes wrong, I need to go through the scene detection part again, which could easily be avoided if I could generate a .csv with timeframes for each scene just once and then use it for every subsequent attempt.
Thank you for your amazing work.

Breakthrough (approved):
Sorry for the delay in getting to this, looks like a great starting point.  Will get this merged in and moved up to the v0.6.2 release as soon as possible.

Thanks so much for all your help!

Issue #53, opened 2018-05-07
Color histogram method


r1b:
I noticed that you have a stub for a `HistogramDetector`. I have completed a proof-of-concept using this method in [this notebook](https://github.com/r1b/scene-detection/blob/master/ColorHistogram.ipynb).

Do you think this is a sound approach? I don't have a lot of experience with this kind of problem so I am curious to hear your thoughts. I would love to integrate this into pyscenedetect.

Breakthrough:
Very interesting approach.  Have you given some thought to benchmarking it versus Albanie's shot detection benchmarks to see how it performs?

It would be quite interesting to have some additional detection techniques in PySceneDetect for sure.  I haven't had much time to keep up with the project lately, but I'm always open to pull requests!

Breakthrough:
Also @r1b, this definitely does seem to be a sound approach - there is plenty of literature referencing this method.

I will have to do some more research into your work, but definitely plan to get started on something like this in the future.  I have added this to the backlog of issues, indicating it will be worked on for the next major release of PySceneDetect following v0.5 (the project is almost done a heavy refactor which should make developing detectors much easier).

Also just curious, what is your development environment like?  I'm not too familiar with Python notebooks, but some of the analysis (esp. with regards to the graphs) would be immensely useful.  Just curious what you're using, if it's an interactive environment.

r1b:
Wonderful! When I have some free time I will play with it some more. I am using [Jupyter](http://jupyter.org/) as a development environment to take advantage of the inline plotting & exploratory workflow. Usually I use `matplotlib` for plots but here I use [Seaborn](https://seaborn.pydata.org/) mostly just to use `distplot` (but I also think they look nicer).

Breakthrough:
Thank you for the references, will definitely be checking out Jupyter. One thing I'm experimenting with is automatic threshold detection for the content/threshold detectors, and had a few high-level questions if you don't mind me picking your brain.

I noticed in your notebook that you're using the equation for threshold T = μ + ασ, where μ is the mean, σ is the standard deviation, and α is a constant set to 5.  Just curious, how did you arrive at this value?  Does it need to be adjusted for different types of source material?

For the histogram itself, have you experimented at all with changing the number of frequency bins in the histograms, or generating the colour code in HSV space instead of RGB?  Just trying to flesh out the design by trying to get a feel for what options need to be presented to the user when calling `detect-histogram`, and what the exact equations/algorithms need to be in place.

[...]

Issue #100, opened 2019-04-14
Detect faded scene changing


llafcode:
Hi @Breakthrough ,

Thank you very much for your awesome software! I'm using it to detect scenes from some cartoon videos, see the original video [here](https://www.dailymotion.com/video/x3bas8y). 

Now I'm having some problems of getting what I want. Specifically:
I want to detect the **real scene changes** (not shot changes) in the video. There are usually two ways of representing **real scene changes**: 

1. fade into black screen and fade into next scene;
2. fade directly into the next scene without a black screen.

The content-detector method works well for shot changes while the threshold-detector works well for **real scene changes** (option 1). But I'm missing many scenes with scene changes in the second way. For example, check scene changes happened at 03:22 to 03:23 and 05:52 to 05:53 in the original video I just posted.

I struggled with adjusting the parameters but I failed to achieve what I want. Any ideas of how I can make it?

Many thanks!

Breakthrough:
Hi @llafcode;

Although it's not documented, multiple detectors are allowed at once (e.g. you can actually specify *both* `detect-content` and `detect-threshold` together).  Have you tried this?  Sounds like this might be a good use case for testing it out.

Thanks for the feedback, let me know if it works out for you.

llafcode:
Hi @Breakthrough,

Thank you for your advice! Sounds interesting! Will try it later. 

If I understand correctly, I can include both detect-threshold and detect-content within one command line, am I right? If so, how should I specify different threshold for the two detectors? By using two -t options? Sounds weird...

Breakthrough:
@llafcode correct, that is possible.  The output scene list will be the combination of scenes detected by both detectors.

llafcode:
@Breakthrough Hi, I tried your suggestion. Although it didn't work exactly as I want, I still got better results. 
I also checked the states files, and see both rgb and hsv states at the timestamp I wanted to be cut. The result is I don't think the current two algorithm can cut in the way I want. 
But since no splitting program is perfect, your tool is still one of the best I found. 
Thank you again!

Breakthrough:
Hi @llafcode;

I just re-analyzed the video, and sorry, I misinterpreted your original question.  Indeed PySceneDetect currently doesn't deal well with slow fades like you mentioned.

If anyone does happen to have suggestions on how to approach the problem, I would be most welcome to investigate the possibility of adding support for it to PySceneDetect.

Thank you!

allenday:
[...]

PR #295, opened 2022-10-27
Histogram detector

wjs018:
As discussed in #53 by @r1b, I have taken inspiration from his notebook as well as [this paper](https://www-nlpir.nist.gov/projects/tvpubs/tvpapers03/ramonlull.paper.pdf) (pdf) and implemented a color histogram based detector. Command syntax is as follows:

```
    Command: 
        detect-hist
    
    Arguments:
        --threshold, -t
            Threshold (float) that must be exceeded to trigger a cut.
        
        --bits, -b
            The number of most significant bits to keep when quantizing the color
            channels of each frame.
        
        --min-scene-len, -m
            Same as other detectors
```

For some detail on the detection algorithm, it works in a couple steps:

1. The frame of the video is separated by color channel and each color channel is quantized. The quantization is done by only retaining the `--bits`/`-b` most significant bits of each pixel. The purpose of this step is to reduce the size of the histogram we will be generating in a future step to make it a bit more tractable.
2. After quantizing each color channel, we want to combine the bits for each pixel of each color channel into one array. For example, with `--bits 2`, the resulting array would have elements with bit values of `0bRRGGBB` where RR are the two most significant bits from the red channel, GG from the green channel, and BB from the blue channel. With `--bits 4` this becomes `0bRRRRGGGGBBBB` and so on. This is done by performing bitwise shift operations on the different color channels before joining them using bitwise or operations.
3. With this one array, a histogram is calculated. The number of bins is given by the possible number of colors after all the quantization. So, for `--bits 4` that would be 2<sup>12</sup> (12 comes from the 4 bits for each of the three color channels). Using `--bits 2` would be 2<sup>6</sup> bins.
4. This histogram is subtracted element-wise from the previous frame's histogram, and the resulting array has all its elements' absolute values summed to find the total difference. This is the value that is checked to trigger cuts and is what is recorded in the stats file as `hist_diff`.

There are a couple things that I should mention with the current implementation. First and foremost is that the input frames need to be an 8-bit color image. This means that grayscale images that don't have the three color channels will not work. Also, inputs that have a bit-depth greater than 8 bits such as image sequences of 16-bit images will not work. I have included some checks to make sure the input is of the right shape and dtype.

The threshold is very sensitive to changes in the analysis parameters. Changing options like the downscale factor or the `--bits` value will have a large impact on what a good threshold would be. Just something to note as the other detectors are not as sensitive to these kind of changes. For defaults, I have chosen values that work well on the goldeneye clip with default downscaling.

Computationally, this detector is not as efficient as others. I have done some testing on my machine and included below the average fps of the detection for different detection algorithms with both default downscaling and `-d 1`.

| detector           | fps with `-d` default | fps with `-d 1` | Other Notes |
|--------------------|-----------------------|-----------------|-------------|
| `detect-hist`      | 637.78                | 49.31           | `-b 4`      |
| `detect-hist`      | 1450.72               | 53.43           | `-b 2`      |
| `detect-content`   | 1422.79               | 300.61          |             |
| `detect-adaptive`  | 1480.52               | 293.96          |             |
| `detect-threshold` | 1425.77               | 1228.24         |             |

If this is something worth cleaning up/improving, I can work on tests and docs.


Breakthrough (approved):
Thanks so much for taking this on, great work.  Looking forward to experimenting to see what performance and accuracy improvements we can make for this.

Breakthrough (dismissed):

Breakthrough (approved):

Breakthrough:
Sorry for the commit spam, I was trying to use the web editor to get this back up to date on the latest changes in develop.  There's still a few failing tests, but this has been merged into develop to continue work on it.  I'll fix the failing tests up as soon as possible and we can continue figuring out the remainder on #53 .

elexor:
Would it be possible to have a mode that only detects hardcuts ignoring gradual transitions? 

It would be useful for video frame interpolation, where the goal is to stop interpolation over cuts but not overdetect continuous motion.

Breakthrough:
@elexor See #100 for an open issue regarding fade detection.  Right now PySceneDetect only has the ability to detect fast cuts (detectors can **only** output hard cuts).

When using the threshold detector, a cut is placed between the fade in/fade out point.  API support for non-hard cuts is planned in the future, which would allow for better handling of cases like this.

PR #324, opened 2023-05-25
Update similar.md

Mtillmann:
Added Chaptertool to the list of similar tools. It can be used to integrate PySceneDetect output into more complex workflows


Breakthrough (approved):

Breakthrough:
Thanks for adding this, interesting project!  My apologies for the way the scene output format is right now - it was originally intended for human readability and not automated processing.  Eventually I want PySceneDetect to support a more standardized format for scenes, but this is important to know for making any future breaking changes.

There's still a lot of work to be done with this project in regards to the output of scene metadata  If you have any opinions, I would be happy to hear any feedback or recommendations you have for the project - especially in regards to the format of scene/chapter lists like this.

Mtillmann:
Hi, thanks for accepting the PR. I do not have a strong opinion on how a good format could look. After working with so many of them I think that they're all created to satisfy a certain requirement at the time they were created. 

However, if it was up to me, I think I'd go with WebVTT: 
- has an actual specification
- very easy to read by humans
- easy to parse and write
- works in the browser (for example to inspect detection result) 
- supports more features like subtitles

Cheers

PR #358, opened 2023-10-13
[docs] Fix typos

kianmeng:
Found via `codespell -S *.aip,*.enc -L datas,bu,te`

### DIFF
--- a/dist/windows/LICENSE-PYTHON
+++ b/dist/windows/LICENSE-PYTHON
@@ -414,7 +414,7 @@ bzip2/libbzip2 version 1.0.6 of 6 September 2010
  * The implementation was written so as to conform with Netscapes SSL.
  *
  * This library is free for commercial and non-commercial use as long as
- * the following conditions are aheared to.  The following conditions
+ * the following conditions are adhered to.  The following conditions
  * apply to all code found in this distribution, be it the RC4, RSA,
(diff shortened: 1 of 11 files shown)


wjs018 (commented):
All of these typo corrections look accurate. I would approve, but this is making me realize that the python license included in PySceneDetect is pretty old as none of these sections are present in the current python license. 

@Breakthrough would you rather update the license and then merge the rest of the corrections?

dist/windows/LICENSE-PYTHON
Breakthrough, line 1: Please revert the changes to this file, this file needs to be reproduced verbatim for this project.

Breakthrough:
Thanks for fixing this up.  @wjs018 I think the license is correct, but I can double check for the next release.  The `dist/windows/LICENSE-PYTHON` should be similar to the LICENSE.txt file on the Python installed on your system.  Is that the same one you were referring to?

kianmeng:
@Breakthrough Done. Please check.

wjs018 (approved):

wjs018:
@Breakthrough I was comparing to the python license [here](https://docs.python.org/3/license.html). There are probably version differences.

Breakthrough (approved):

kianmeng:
:partying_face: :partying_face: :partying_face: :partying_face: :partying_face:

PR #418, opened 2024-09-04
Transition from yapf to ruff

Breakthrough:
This PR switches the project to use [`ruff`](https://docs.astral.sh/ruff/) for the project.  It seems to be much faster and more configurable, so I think this is a nice quality of life improvement.


wjs018 (dismissed):

wjs018:
I am going to have to get this set up on my end, but it looks like there is a nice native way to do this in VSCode and all the formatting options are set in pyproject.toml. So, I don't think this will be too problematic.

Issue #388, opened 2024-04-10
QP file output


tormento:
x265 (and other software) accepts a .qp file to force a frame type to be applied to a certain frame.

An example is:
```
0 I -1
1640 I -1
3789 I -1
17191 I -1
31784 I -1
33927 I -1
```
to have proper scene change cuts.

[Here](https://x265.readthedocs.io/en/master/cli.html?highlight=qpfile#cmdoption-qpfile) you can find the syntax.

It can be derived by chapters but it's almost of no use when wanting a clean cut frame per each scene.

Would you mind to add that kind of output?

Breakthrough:
Thanks, this falls under some of the other open issues to expand on output formats.  I think we should add a new command like `export-scenes` that is dedicated for this purpose, so that the cuts can be saved in various formats.  Updated #344 to include a link to this issue.

I would imagine the command to look something like `export-scenes -o "cuts.qp"` and have the format auto-detected from extension where possible, or explicitly specified with another flag.  We could also allow specifying `export-scenes` several times to generate the output in different formats.

tormento:
Thanks, do you have any ETA?

Breakthrough:
Unfortunately no ETA at this time, but in v0.6.3 there are some features that make workarounds for this a bit easier.  You can change [the format cuts are generated with](https://github.com/Breakthrough/PySceneDetect/blob/releases/0.6.3/scenedetect.cfg#L230) in a config file, and then convert the resulting CSV file or command line output to the QP file.

If in your config file you use:

    [list-scenes]
    cut-format = frames
    
And then use the list-scenes command to generate a CSV file, the first row will be the frames that you need to add `I` frames for.  If you're using the Python API, the `detect` function should already return the info you need.

tormento:
Tried your release, sorry for the delay.

It works ok but the generated CSV is not properly "standard".

I mean: looking at the file, we have a first part with Timecode list and a second part with Scene Number, that are CSV formatted in a really different way each other.

Please, write two files so that I can directly import the Scene number and frame in Excel.

When you will have time (ifyou want to ease my CSV file "postprocessing" :) ), simply output a file with a column of plain scene start frames, almost such as the example I did in the first post.

Breakthrough:
@tormento if you [use the `--skip-cuts` option](https://www.scenedetect.com/docs/latest/cli.html#cmdoption-scenedetect-list-scenes-s) the file will omit the first row to allow it to be imported directly into Excel.

[...]

Issue #323, opened 2023-05-16
Export scenes in chapter format


schtritoff:
**Description of Problem & Solution**
As per project [roadmap](https://github.com/Breakthrough/PySceneDetect/blob/master/docs/features.md#version-roadmap), I'm opening this feature request to have an option to export scenes in interchangeable/standard format such as ffmetadata, nero/vorbis, etc. There is also a similar feature request but specific for Final Cut XML format https://github.com/Breakthrough/PySceneDetect/issues/156. 
Some common chapter formats with examples are listed here https://github.com/rigaya/NVEnc/blob/master/NVEncC_Options.en.md#--chapter-string

**Proposed Implementation:**
This should be a parameter to [list-scenes](https://github.com/Breakthrough/PySceneDetect/blob/2e1000fc0e7a3ebb1169266dbf98135840d8cb20/manual/cli/commands.rst?plain=1#L154) command as a new option. For example `list-scenes` -e, --export-format csv/nero-vorbis/ffmetadata/matroska/applexml with 'csv' being the same as -s, --skip-cuts switch.

**Alternative Solutions:**
chaptertool can convert PySceneDetect format since few days ago - https://github.com/Mtillmann/chaptertool/issues/1
powershell script doing similar - https://gist.github.com/schtritoff/8e0105060cb07c118863867429473830

Issue #156, opened 2020-03-27
Final Cut Pro X - XML File


cbitterfield:
**Description of Problem & Solution**

Request the ability export scene cut information directly to a FInal Cut Pro XML file.

**Media Examples:**

Guide to XML
https://support.apple.com/guide/final-cut-pro/use-xml-to-transfer-projects-verdbd66ae/mac.

**Proposed Implementation:**

From what I understand it is the same information as the CSV file but in "FCPX" XML format

**Alternative Solutions:**
Writing a python converter between the CSV output and the XML format.

Breakthrough:
I'm not familiar with the XML format, could you share one for an existing video with split scenes so I could judge the effort required?  Unfortunately the page you linked to does not provide any additional information about the format.

Thank you!

Breakthrough:
Marking as on-hold until further information can be gleaned as to the scope of this feature.  Closing issue to cleanup backlog, feel free to re-open in future if desired.

camjac251:
I am very interested in a feature like this. Having it could save a lot of time with my workflow and also hopefully fix a few quirks with the current one (like with colorspace conversion or disk space required for lossless conversions)

camjac251:
I tried to find if someone else might've made something like this before but could only find https://github.com/Mirellanmawl/PySceneDetect-CSV-to-Vegas-EDL and https://github.com/duolanda/scene-detect-to-xml
The second one I believe takes the final clip video files generated by PySceneDetect and creates an FCP XML from that, instead of using the CSV file as the input.

Breakthrough:
If you can provide an example of what this XML should look like I can certainly look into adding support for it.  I don't have Final Cut Pro so I will need some assistance with testing as well.

camjac251:
I can work on that, to get one to you. I use Premiere Pro and Davinci Resolve. FCP XML is pretty much supported with all major NLEs luckily and has been better for me than EDL. You could save one out or import one in both programs.
It might take me a little while to create a test example but maybe this might be more helpful https://developer.apple.com/documentation/professional_video_applications/fcpxml_reference
https://github.com/Cubxity/ResolveParser
or the work that this user has done https://github.com/duolanda/scene-detect-to-xml/blob/main/xml_creator.py
or https://github.com/ldsxp/fcpxml/blob/master/fcpxml/fcp.py

Breakthrough:
[...]

PR #426, opened 2024-09-28
[cli] Move output command state out of CliContext

Breakthrough:
This PR makes it a lot easier to integrate new output commands into the CLI.  The goal is to have the controller treat all output commands the same by operating on the result of the processing pipeline.  For dependent jobs, they can still use the shared context to communicate results.  This has a few benefits:

- **much** easier to add new commands that operate on the result of scene detection and far less code
- can now specify output commands multiple times, e.g. can run `split-video` twice to generate output videos with different parameters
- logging parameters is more consistent now with `--verbosity debug`

This should also help reduce the friction to adding new output formats, supporting #156, #323, and #388.  As part of this change, some log statements will now be emitted later when a command is actually running.  Previously, some commands would emit things like the location of saved files when processing the command-line arguments, rather than when the sub-command was about to run.

### DIFF
--- a/scenedetect/__main__.py
+++ b/scenedetect/__main__.py
@@ -22,10 +22,10 @@ from scenedetect.platform import FakeTqdmLoggingRedirect, logging_redirect_tqdm
 
 def main():
     """PySceneDetect command-line interface (CLI) entry point."""
-    cli_ctx = CliContext()
+    context = CliContext()
     try:
         # Process command line arguments and subcommands to initialize the context.
-        scenedetect.main(obj=cli_ctx)  # Parse CLI arguments with registered callbacks.
+        scenedetect.main(obj=context)  # Parse CLI arguments with registered callbacks.
     except SystemExit as exit:
         help_command = any(arg in sys.argv for arg in ["-h", "--help"])
         if help_command or exit.code != 0:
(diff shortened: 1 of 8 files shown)


wjs018:
Alright, I have had a good look through this and I think I have wrapped my head around it. I don't use the cli too often, so I always need to refresh my memory on how it all works. If I understand what this is doing correctly, then it works like this under this paradigm:

1. `scenedetect/__main__.py` calls `run_scenedetect` from `scenedetect/_cli/controller.py`
2. The cli is parsed by `scenedetect/_cli/__init__.py` and any commands that are output commands are simply added to `context.commands` with named arguments for later use
3. The `run_scenedetect` function then proceeds to do the detection (or loads the scenes)
4. Within `run_scenedetect` the `_detect` function returns the detected scenes and cut list
5. The list of `context.commands` are then looped through with the scenes and cut list passed to each command

I think this makes sense and does make it easier to add additional output formats in the future. To add a new output format, I think the changes that would need to be made would be:

- The cli command would need to be added to `scenedetect/_cli/__init__.py`, adding the command to `context.commands` with the proper args
- The logic of the output processing/formatting would need to be defined in `scenedetect/_cli/commands.py`. This could be pretty a pretty barebones reference to a separate file that does the heavy lifing (like what is done for `export-html` or `save-images`)

wjs018 (approved):

Breakthrough:
Though still WIP, feel free to check out #427 for a concrete example of how this simplifies adding new CLI commands.

PR #427, opened 2024-09-30
[cli] Add new `export-qp` command

Breakthrough:
Add ability to export scene cuts in `qp` format with a new `export-qp` command.

Open Tasks
--------------------

- [ ] Verify resulting `qp` file in test case with a golden file


Breakthrough:
Replaced by #448

PR #438, opened 2024-10-15
Standardize min scene length message across detectors

Janscode:
I found the min_scene_length docs to be a little confusing for the AdaptiveDetector class, I think it's helpful to know 1) the value is in frames and 2) scenes won't be filtered out / dropped, instead they are merged


Breakthrough:
Thanks for improving this.

Breakthrough (approved):

PR #446, opened 2024-10-19
[dist] Update Windows Python/deps + Include MoviePy

Breakthrough:
Updates:

 - Python 3.9 -> 3.13
 - PyAV 10 -> 13.1.0
 - OpenCV 4.10.0.82 -> 4.10.0.84
 - MoviePy


Breakthrough:
Looks like this approach only increases the total distribution size by a few megabytes which is acceptable.

PR #459, opened 2024-11-19
[detectors] Add Koala-36M Based Detector

### LINKED ISSUE #441: Determine Feasibility of Koala-36M method for detector
[Koala-36M](https://koala36m.github.io/) proposes a significantly improved model for scene transition detection (paper: [HTML](https://arxiv.org/html/2410.08260v1) or [PDF](https://arxiv.org/pdf/2410.08260v1))

See [section 4.1](https://arxiv.org/html/2410.08260v1#S4) which uses an SVM classifier.  The performance degredation is just over 2x slowdown, however the accuracy, precision, and recall show marked improvements across the board that likely warrant this change for the majority of users.

Breakthrough:
Work in progress. #441

Based off of https://koala36m.github.io/ (will update this with proper citation when ready for review)

### DIFF
--- a/benchmarks/README.md
+++ b/benchmarks/README.md
@@ -37,6 +37,7 @@ The following results indicate that ContentDetector achieves the highest perform
 |    HashDetector   |  92.96 |   76.27   | 83.79 |         16.26         |
 | HistogramDetector |  90.55 |   72.76   | 80.68 |         16.13         |
 | ThresholdDetector |  0.00  |   0.00    |  0.00 |         18.95         |
+|   KoalaDetector   |  86.83 |   78.38   | 82.39 |         97.75         |
 
 ## Citation
 ### BBC
--- a/benchmarks/bbc_dataset.py
+++ b/benchmarks/bbc_dataset.py
@@ -12,15 +12,19 @@ class BBCDataset:
 
     def __init__(self, dataset_dir: str):
         self._video_files = [
-            file for file in sorted(glob.glob(os.path.join(dataset_dir, "videos", "*.mp4")))
+            file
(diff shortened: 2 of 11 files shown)


benchmarks/README.md
```diff
@@ -37,6 +37,7 @@ The following results indicate that ContentDetector achieves the highest perform
 |    HashDetector   |  92.96 |   76.27   | 83.79 |         16.26         |
 | HistogramDetector |  90.55 |   72.76   | 80.68 |         16.13         |
 | ThresholdDetector |  0.00  |   0.00    |  0.00 |         18.95         |
+|   KoalaDetector   |  86.83 |   78.38   | 82.39 |         97.75         |
```
awkrail, line 40: Oh, interesting. Even though Koala detector uses ML-based methods, it does not achieve good performance compared with rule-based detectors?

Breakthrough: It's possible I have a bug somewhere in the implementation.  This was a first-pass at implementing this with just a single video test case.

I haven't reviewed the code in some time, so if you spot anything that looks wrong, please do let me know!  I haven't had time to revisit this PR in a while.  If you want to develop this detector further, I would be happy to clean this up as-is so others can make modifications.

awkrail: I see. To be honest, I am not sure about the performance of Koala-37M because the authors do not evaluate the shot detection accuracy in the paper. Because they released the code (https://github.com/KwaiVGI/Koala-36M/blob/main/trainsition_detect/VideoTransitionAnalyzer.py), let me check the performance on BBC and AutoShot. I will report it in this thread.

PR #484, opened 2025-02-14
[benchmark/WIP] Benchmarking pyscenedetect detectors' performance

awkrail:
Discussed in #481 
This PR aims to evaluating PySceneDetect detectors' performance in terms of both latency and accuracy.
The accuracy measurement is based on [LoveU Challenge Track 1](https://sites.google.com/view/loveucvpr21/track-1), which calculates recall, precision, and F1 scores. The following is quoted from the task description:
```
- We use Relative Distance (Rel.Dis) to determine the correctness of each prediction. Rel.Dis is the error between detected and ground-truth timestamps, divided by the length of the whole video. Given a fixed threshold for Rel.Dis, we can determine whether a detection is correct (i.e. <=threshold) or incorrect (i.e. >threshold), then compute precision, recall and F1 score on the whole dataset.
- Note that for each video, we multiple raters who make annotations independently. We (1) compare our detection result with each rater’s annotation and (2) select GT for this video as the rater’s annotation which leads to the best F1 score among all raters.
- The official metric used in this task is F1@5%, which is defined as the F1 score computed with threshold 5%.
```
**Question**: F1@5% is OK? I think that 5% difference is critical, so probably changing threshold to <1% may be better.

I will use two datasets: RAI and BBC. Both datasets are uploaded on Zenodo. Please check README.md in the `benchmark/` directory to download them.
- [RAI](https://zenodo.org/records/14865179)
- [BBC](https://zenodo.org/records/14865504)

PR #486, opened 2025-02-19
[benchmark] evaluate the detectors on the AutoShot dataset

awkrail:
Related to #484. In addition to BBC, I implemented benchmarking codes on the AutoShot dataset. The original work uses 200 videos for the test set, yet 36 videos are missing ([link](https://github.com/wentaozhu/AutoShot/issues/6)). Hence, I evaluated the codes on the remaining videos.

**Results** 
The following table shows the results, indicating that AdaptiveDetector achieves the highest scores. This is consistent with BBC.

|      Detector     | Recall | Precision |   F1  | Elapsed time (second) |
|:-----------------:|:------:|:---------:|:-----:|:---------------------:|
|  AdaptiveDetector |  70.77 |   77.65   | 74.05 |          1.23         |
|  ContentDetector  |  63.67 |   76.40   | 69.46 |          1.21         |
|    HashDetector   |  56.66 |   76.35   | 65.05 |          1.16         |
| HistogramDetector |  63.36 |   53.34   | 57.92 |          1.23         |
| ThresholdDetector |  0.75  |   38.64   |  1.47 |          1.24         |

### DIFF
--- a/benchmarks/bbc_dataset.py
+++ b/benchmarks/bbc_dataset.py
@@ -20,7 +20,7 @@ class BBCDataset:
         assert len(self._video_files) == len(self._scene_files)
         for video_file, scene_file in zip(self._video_files, self._scene_files):
             video_id = os.path.basename(video_file).replace("bbc_", "").split(".")[0]
-            scene_id = os.path.basename(scene_file).split("_")[0]
+            scene_id = os.path.basename(scene_file).split("-")[0]
             assert video_id == scene_id
 
     def __getitem__(self, index):
--- a/benchmarks/README.md
+++ b/benchmarks/README.md
@@ -16,7 +16,14 @@ unzip BBC/videos.zip -d BBC
 rm -rf BBC/videos.zip
 ```
 
-### Evaluation
+### AutoShot
+Download `AutoShot_test.tar.gz` from [Google drive](https://drive.google.com/file/d/17diRkLlNUUjHDooXdqFUTXYje2-x4Yt6/view?usp=sharing).
+```
+tar -zxvf AutoShot.tar.gz
+rm AutoShot.tar.gz
+```
+
+## Evaluation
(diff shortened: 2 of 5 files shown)


benchmarks/bbc_dataset.py
Breakthrough, line 1: Question: Do you know what the difference between the text files listed under https://zenodo.org/records/14865504 versus the ones in the .zip?

I noticed some cuts at the beginning were different.  I last updated the benchmark results using these ones, so I'll make sure to do that with the ones from the .zip you linked in the README.

Breakthrough (approved):

Breakthrough:
Awesome, thanks for the PR!  Judging by the results for `ThresholdDetector` this dataset probably includes some more variety of transitions like dissolves/fades.

awkrail:
@Breakthrough Yes, and I think that video cut type annotations (e.g., abrupt, dissolves, fade-in/out) may be useful. I will attach it with video cuts in the future for detailed evaluation.

benchmarks/bbc_dataset.py
awkrail: > Question: Do you know what the difference between the text files listed under https://zenodo.org/records/14865504 versus the ones in the .zip?

Yes, they are slightly different.
I checked the original BBC annotation (records/14865504) but the some cut annotations are incorrect. For example, in 01_From_Pole_to_Pole.txt in https://zenodo.org/records/14865504, the first cut finished at the 632 frame, but the 632-th frame is not the end of the cut. The correct cut is 649-th frame. I checked such cases manually, and fixed them by changing them to be `curr_end = next_start - 1`. The number of such incorrect cuts are small, so I think that the results does not change significantly.

Could you update the results based on the new ones?

Breakthrough: Thanks for looking into this.

> Could you update the results based on the new ones?

For sure, I will do that as soon as possible!  :)

PR #489, opened 2025-02-20
Benchmark run as module

Breakthrough:
Allow running the benchmarks from the root of the repo as `python -m benchmark` so that it can work with a local development version of `scenedetect`.


awkrail (approved):
LGTM

Breakthrough:
I've also re-run the benchmarks as discussed in #486.

Issue #401, opened 2024-06-16
Remove downscaling for perceptual hash


Breakthrough:
The HashDetector already resizes the input to a specific size, so we should avoid downscaling the input for this particular detector.  This might involve changing the SceneDetector interface so that detectors can choose if they want a pre-scaled version of the frame, or if they need to work with the original.

The perceptual hash detector is a bit of a special case, since it requires that the input be a square for the DCT. Right now the image is scaled twice, which is not ideal, especially if the downscaled version is smaller than the DCT size (since we're throwing away information).

This doesn't matter too much for accuracy since we apply a low-pass filter on the result, but it is something we should avoid doing for the pipeline.

Breakthrough:
It's probably best to just make downscaling a per-detector option, and allow `SceneManager` to automatically try to apply it if a global one was set.

wjs018:
> make downscaling a per-detector option

This makes sense. It should be pretty easy to do on the python side, but will need more changes on the cli side to parse all those parameters.

Breakthrough:
From my tests it seems this made things worse in general.  Both performance and accuracy were better/worse in some cases, but in general it didn't seem worth the additional complexity.  On most benchmarks the improvement was marginal, and in some cases far worse.

PR #490, opened 2025-02-22
[detectors] Add downscale performance hit #401

Breakthrough:
Adds a new performance hint property to detectors to indicate if they have a significant performance benefit when using a downscaled version of the frame with minimal accuracy loss.

Disable downscaling if set to `auto` for `HashDetector` (`detect-hash`) as it internally resizes the image to a specific size already.

### DIFF
--- a/benchmark/__main__.py
+++ b/benchmark/__main__.py
@@ -1,6 +1,6 @@
 import argparse
-import time
 import os
+import time
 
 from tqdm import tqdm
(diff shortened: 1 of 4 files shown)


Breakthrough:
This actually seems to make things slower, and while it does have marginal improvements on BBC, is less accurate on the AutoShot database.  Will just leave this as-is.

PR #491, opened 2025-02-22
[benchmark] Allow running all benchmarks

Breakthrough:
Add --all flag to run all detector/dataset combinations.

Add and require --detailed flag to output information for each video. Add extra separators in detailed mode to help more easily see the overall result versus each individual video.


awkrail (approved):
LGTM. This feature seems to be very useful!
I think that adding explanations about benchmark to docs is useful. I will do it tomorrow.

PR #498, opened 2025-03-10
[benchmark/WIP] Add ClipShots to the evaluation set

awkrail:
Related to #484. ClipShots is the largest dataset for video shot detection. In this PR, I add it to the evaluation set.


Breakthrough (approved):
Thanks!  How big is the dataset / could you run the benchmarks to fill in the table?

I can also tackle that, but wanted to know if you were planning on doing so first.

awkrail:
> Thanks! How big is the dataset / could you run the benchmarks to fill in the table?

The number of videos (dataset size) is 500, but the annotation styles seem to be different from other datasets, BBC and AutoShots. So, now I am trying to reproduce the performance as with other datasets, yet currently could not. Let me take some time to do it.

> I can also tackle that, but wanted to know if you were planning on doing so first.

I plan to add the dataset, evaluate the performance, and fill the scores in the Table in README.md.

PR #512, opened 2025-06-04
[backends] Add comprehensive test suite for VideoStreamMoviePy

chyang-nycu:
Screenshot 2025-06-05 at 06 18 58


Breakthrough:
I like where this is going in concept, but would prefer to have this abstracted over the backend itself similar to the existing `test_video_stream.py`.  Have you considered adding more tests there to increase code coverage?

The tests in that file are run across all backends to ensure consistency, and there might be some overlap.

Breakthrough:
@chyang-nycu are you still open to merging this?  It just needs to pass `ruff format` to be merged.

Breakthrough (changes requested):

PR #520, opened 2025-09-05
Import numpy in platform.py

Atanu484:
Import NumPy before OpenCV in platform.py to prevent import/segfault issues (FreeBSD & cross-platform safety)

On FreeBSD, importing OpenCV (cv2) before NumPy can blow up with errors like “numpy.core.multiarray failed to import” or even a segfault. The reason is that OpenCV relies on NumPy under the hood, so if NumPy hasn’t been loaded yet, things can go wrong.

FreeBSD Ports Tree already carries a patch that imports numpy to prevent this. Thought it would not be a bad idea to upstream this

https://github.com/freebsd/freebsd-ports/blob/main/multimedia/py-scenedetect/files/patch-scenedetect_platform.py

### DIFF
--- a/scenedetect/platform.py
+++ b/scenedetect/platform.py
@@ -27,6 +27,7 @@ import sys
 import typing as ty
 
 import cv2
+import numpy
 
 ##
 ## tqdm Library


Breakthrough:
Thanks for letting me know, I would be happy to upstream to avoid needing a patch!

Breakthrough (changes requested):
I'm curious, has this caused any actual issues?  We also import `cv2`  in `__init__.py` before this code has run:
https://github.com/Breakthrough/PySceneDetect/blob/ad9d35dd94f8daa2d46a69c484b01d469040c826/scenedetect/__init__.py#L21-L30

That being said, if this does need to be done, I think the patch was applied incorrectly.

PR #551, opened 2026-05-17
Add --expand flag to split-video (#115)

### LINKED ISSUE #115: Allow for clips to be split with content outside of start/end timeframe

charlesvestal:
## Summary

Resolves #115 (which I filed back in 2020).

Adds a `--expand-to-video` flag to `split-video` so that when scene detection is limited to a sub-region of a video via `time -s/-e`, the resulting split clips extend outward to cover content outside the analysis window. The first output clip extends back to the start of the video, and the last clip extends to the end.

This is useful for files with roughly known split points, where you want to target a specific area only — e.g. a 30-minute file where you know there's a cut somewhere between 13:00 and 17:00. You can limit detection to that range (fast, no false positives elsewhere) and still get two 15-minute clips covering the full video rather than two ~2-minute clips of just the analyzed window.

## Implementation

- New pure helper `expand_scenes_to_bounds(scenes, start, end)` in `scene_manager.py` (sits next to `get_scenes_from_cuts`, same docstring style). Unit-tested in isolation.
- CLI flag `--expand-to-video` on `split-video`, with a matching `expand-to-video` config entry (default `False`).
- `split_video` command handler invokes the helper before dispatching to either ffmpeg or mkvmerge, so both splitters get the expanded list. Falls back to the original scene boundaries with a warning if `video_stream.duration` is `None`.

## Disclosure

I used Claude to help draft this. All code was reviewed and verified by me; the smoke testing below was driven from my end against synthesized fixtures.

## Test plan

- [x] `pytest tests/test_scene_manager.py` — 10/10 pass (6 existing + 4 new unit tests for `expand_scenes_to_bounds`)
- [x] `pytest tests/test_cli.py` — 49 pass, 2 unrelated env failures (missing `pyav`/`moviepy` deps locally)
- [x] `ruff check` and `ruff format --check` clean on all touched files
- [x] End-to-end smoke test on a synthesized 30s, 6-segment color test video (cuts every 5s):
    - **ffmpeg path** with `time -s 12 -e 22` + `--expand-to-video` → 3 clips of ~15s / ~5s / ~10s (keyframe-rounded), first-frame colors red / yellow / magenta
    - **mkvmerge path** with the same args → 3 clips of exactly 15.0s / 5.0s / 10.0s, same red / yellow / magenta colors
    - No `time` window + `--expand-to-video` → no-op, full coverage preserved
    - `time -s 12` only + `--expand-to-video` → first clip extends back to 0
    - `time -e 22` only + `--expand-to-video` → last clip extends to video end

### DIFF
--- a/scenedetect/_cli/__init__.py
+++ b/scenedetect/_cli/__init__.py
@@ -1295,12 +1295,12 @@ Customized filenames:
     ),
 )
 @click.option(
-    "--expand-to-video",
+    "--expand",
     is_flag=True,
     flag_value=True,
     default=False,
     help="Extend the first/last output clips to cover the full input video, even if `time -s/-e` limited the analysis window. Useful for keeping content outside the analyzed region attached to the adjacent split.{}".format(
-        USER_CONFIG.get_help_string("split-video", "expand-to-video")
+        USER_CONFIG.get_help_string("split-video", "expand")
     ),
 )
 @click.pass_context
@@ -1315,7 +1315,7 @@ def split_video_command(
     preset: str | None,
     args: str | None,
     mkvmerge: bool,
-    expand_to_video: bool,
+    expand: bool,
 ):
     ctx = ctx.obj
     assert isinstance(ctx, CliContext)
@@ -1382,7 +1382,7 @@ def split_video_command(
         "output": ctx.config.get_value("split-video", "output", output),
         "show_output": not ctx.config.get_value("split-video", "quiet", quiet),
         "ffmpeg_args": args,
-        "expand_to_video": ctx.config.get_value("split-video", "expand-to-video", expand_to_video),
+        "expand": ctx.config.get_value("split-video", "expand", expand),
     }
     ctx.add_command(cli_commands.split_video, split_video_args)
 
--- a/scenedetect/_cli/__init__.py
+++ b/scenedetect/_cli/__init__.py
@@ -1294,6 +1294,15 @@ Customized filenames:
         USER_CONFIG.get_help_string("split-video", "mkvmerge")
     ),
 )
+@click.option(
+    "--expand-to-video",
+    is_flag=True,
+    flag_value=True,
+    default=False,
+    help="Extend the first/last output clips to cover the full input video, even if `time -s/-e` limited the analysis window. Useful for keeping content outside the analyzed region attached to the adjacent split.{}".format(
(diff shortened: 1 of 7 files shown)


Breakthrough:
Thanks for taking a pass at this!  Appreciate the up-front disclosures as well.  I'll get the main branch ready to continue development after the 0.7 release now (next will be 0.7.1).  Happy to include this feature in the next release.

scenedetect/_cli/__init__.py
```diff
@@ -1294,6 +1294,15 @@ def list_scenes_command(
         USER_CONFIG.get_help_string("split-video", "mkvmerge")
     ),
 )
+@click.option(
+    "--expand-to-video",
```
Breakthrough: Optional: We can just use `--expand` here if you want. `--expand-to-video` is fine too, but happy to use a more terse name as well.

scenedetect/_cli/config.py
```diff
@@ -468,6 +468,7 @@ class FcpFormat(Enum):
     "split-video": {
         "args": _DEFAULT_FFMPEG_ARGS,
         "copy": False,
+        "expand-to-video": False,
```
Breakthrough: Can you update the `scenedetect.cfg` template in the root of the repo as well to include the new config option?

Breakthrough (changes requested):
Thanks for the PR!  Is it possible for you to update the CLI docs as well to include the new flag?  Feel free to defer if you want, just say "prefer if you can update docs" and I'll take it from there.  Cheers.

Breakthrough (approved):

Breakthrough:
Thanks for the PR!

PR #556, opened 2026-07-03
Remove unused imports and fix re-exports (F401)

Mubashir78:
Remove unused imports across the codebase flagged by Ruff rule F401. Uses explicit re-exports in __init__.py to preserve the public API.


Breakthrough:
Thanks for the PR. Let me think about this for a bit - I'm not sure if I want to commit to the duplicate `from x import y as y` style, or if we should use `__all__` instead to reduce duplication.

Mubashir78:
In my opinion, good practice is to be explicit about imports, and making sure those imports are used. Lines of code does not matter if it can make the parts clearer. To simply use `__all__` is an abstraction which would be confusing to other people in the long run.

scenedetect/backends/__init__.py
```diff
@@ -84,7 +84,7 @@
 #  - Nvidia VPF: https://developer.nvidia.com/blog/vpf-hardware-accelerated-video-processing-framework-in-python/
 
 # OpenCV must be available at minimum.
-from scenedetect.backends.opencv import VideoCaptureAdapter, VideoStreamCv2
+from scenedetect.backends.opencv import VideoCaptureAdapter as VideoCaptureAdapter, VideoStreamCv2
```
Breakthrough: For consistency let's export all of the backends publically here too (e.g. VideoStreamCv2 as VideoStreamCv2, same w/ others below).

Breakthrough (changes requested):

Mubashir78:
Done, added re-exports for VideoStreamCv2, VideoStreamAv, and VideoStreamMoviePy.

Breakthrough:
Last ask, can you rebase + update to export the new items I added for v0.7.1?  Thanks for the PR BTW.

Mubashir78:
Rebased onto latest upstream main and added as re-exports for SourceSpan and VideoStreamConcat (the new v0.7.1 items). No problem!

Mubashir78:
Just wanted to gently follow up on this. Everything should be up to date with the latest v0.7.1 additions. No rush at all, but when you get a chance, would love to hear your thoughts whenever it feels right. Happy to make any adjustments.

Breakthrough (approved):

Breakthrough (approved):
You'll need to run the formatter before submitting, but approved. Let me know when the builds/analyzers are passingl. Thanks for the PR.

Mubashir78:
I ran `ruff --check` & `ruff --fix` across the branch. Hopefully CI is green now.

Breakthrough:
Thanks for the PR!

Mubashir78:
No problem! I will be on the lookout for more code fixes 👍

Cheers mate ✌️

Issue #561, opened 2026-07-26
Scene list has invalid end time for variable frame rate DVDs


nathanmdan:
**Description:**

When generating a scene list from a variable frame rate DVD source (MPEG2 stream in MKV container), the end time of the last shot in the list is earlier than its start time. This results in an error when using split-video with the --mkvmerge option: `Error: Invalid end time for '--split' in [...]. The end time must be bigger than the start time.`, followed by `Error splitting video (mkvmerge returned 2).`, and no files are written.

Interestingly, the invalid end time is the same as the last position of the detection progress bar, which stopped updating at 80% even though it successfully detected 100% of the cuts in the video. This seems to suggest a connection.

Tested with 2 separate MKV-wrapped MPEG2 video files and with different detection methods. All produce the same invalid end time.

Please let me know if any further info would be helpful. I am also interested in looking further into this myself and can follow up if I find the source of the issue.

**Commands:**

`scenedetect -i .\dvd_ntsc_vfr.mkv list-scenes -n`

`scenedetect -i .\dvd_ntsc_vfr.mkv list-scenes -n split-video --mkvmerge`

**Output:**

```
[PySceneDetect] Detecting scenes...
  Detected: 281 | Progress:  80%|█████████████████████████████████▋        | 35840/44745 [00:31<00:07, 1133.57frames/s]
[PySceneDetect] Processed 35840 frames in 31.6 seconds (average 1133.51 FPS).
[PySceneDetect] Detected 282 scenes, average shot length 5.3 seconds.
[PySceneDetect] Scene List:
-----------------------------------------------------------------------
 | Scene # | Start Frame |  Start Time  |  End Frame  |   End Time   |
-----------------------------------------------------------------------
 |      1  |           1 | 00:00:00.000 |          37 | 00:00:01.218 |
 |      2  |          38 | 00:00:01.218 |         751 | 00:00:25.075 |
 
 [...]

 |    280  |       44533 | 00:24:45.884 |       44628 | 00:24:49.104 |
 |    281  |       44629 | 00:24:49.104 |       44712 | 00:24:51.890 |
 |    282  |       44713 | 00:24:51.890 |       35840 | 00:19:55.861 |
-----------------------------------------------------------------------
[PySceneDetect] Comma-separated timecode list:
  00:00:01.218,00:00:25.075, [...] ,00:24:49.104,00:24:51.890
[...]

PR #562, opened 2026-08-04
[bugfix] Fix progress bar accuracy for VFR videos

nathanmdan:
- Addresses progress bar behavior mentioned in Issue #561 where progress bar only goes up to ~80% with VFR videos, even with PyAV backend.

- Progress bar now goes up to 100% for both PyAV and OpenCV backends.

- Please note that this PR does not address the main topic of Issue #561 (Invalid end time in scene lists for VFR videos) which was determined to be unique to the OpenCV backend and is not present when using PyAV.

### DIFF
--- a/scenedetect/scene_manager.py
+++ b/scenedetect/scene_manager.py
@@ -571,6 +571,7 @@ class SceneManager:
         )
         decode_thread.start()
         frame_im = None
+        prev_position = None
 
         logger.info("Detecting scenes...")
         try:
@@ -581,13 +582,22 @@ class SceneManager:
                 if next_frame is not None:
                     frame_im = next_frame
                 assert frame_im is not None
+                if prev_position is None:
+                    prev_position = self._start_pos
+                assert prev_position is not None
                 new_cuts = self._process_frame(position, frame_im, callback)
                 if progress_bar is not None:
                     if new_cuts:
                         progress_bar.set_description(
                             PROGRESS_BAR_DESCRIPTION % len(self._cutting_list), refresh=False
                         )
-                    progress_bar.update(1 + frame_skip)
+                    # Increment progress bar by delta of position.frame_num instead of 1
+                    # to handle VFR video where frame count is an approximation.
+                    # First frame increments by 1 as base case.
+                    delta_pos = position.frame_num - prev_position.frame_num
+                    is_first_frame = position.frame_num == 0
+                    progress_bar.update((1 if is_first_frame else delta_pos) + frame_skip)
+                    prev_position = position
         finally:
             if progress_bar is not None:
                 progress_bar.set_description(
--- a/scenedetect/scene_manager.py
+++ b/scenedetect/scene_manager.py
@@ -582,9 +582,6 @@ class SceneManager:
                 if next_frame is not None:
                     frame_im = next_frame
                 assert frame_im is not None
-                if prev_position is None:
-                    prev_position = self._start_pos
-                assert prev_position is not None
                 new_cuts = self._process_frame(position, frame_im, callback)
                 if progress_bar is not None:
                     if new_cuts:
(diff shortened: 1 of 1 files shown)


scenedetect/scene_manager.py
```diff
@@ -581,13 +582,22 @@ def detect_scenes(
...
                         progress_bar.set_description(
                             PROGRESS_BAR_DESCRIPTION % len(self._cutting_list), refresh=False
                         )
-                    progress_bar.update(1 + frame_skip)
+                    # Increment progress bar by delta of position.frame_num instead of 1
+                    # to handle VFR video where frame count is an approximation.
+                    # First frame increments by 1 as base case.
+                    delta_pos = position.frame_num - prev_position.frame_num
+                    is_first_frame = position.frame_num == 0
+                    progress_bar.update((1 if is_first_frame else delta_pos) + frame_skip)
```
Breakthrough: When `frame_skip > 0`, the skipped frames are already reflected in `position.frame_num`. See the prev. comment for an alternative.

The decode thread queues a frame, then skips, so consecutive queued positions differ by `frame_skip + 1`). Adding `+ frame_skip` on top double-counts them, e.g. with `frame_skip=1` the bar advances 3 per processed frame instead of 2.

scenedetect/scene_manager.py
```diff
@@ -581,13 +582,22 @@ def detect_scenes(
...
                     if new_cuts:
                         progress_bar.set_description(
                             PROGRESS_BAR_DESCRIPTION % len(self._cutting_list), refresh=False
                         )
-                    progress_bar.update(1 + frame_skip)
+                    # Increment progress bar by delta of position.frame_num instead of 1
+                    # to handle VFR video where frame count is an approximation.
+                    # First frame increments by 1 as base case.
+                    delta_pos = position.frame_num - prev_position.frame_num
+                    is_first_frame = position.frame_num == 0
```
Breakthrough: `position.frame_num == 0` won't hold when detection starts from a seek (e.g. `start_time` is set) leading to an off by 1 error in some cases. Both this and the `frame_skip` issue below can be fixed if we keep `prev_position = None` and computing:

```python
delta = 1 if prev_position is None else position.frame_num - prev_position.frame_num
progress_bar.update(delta)
prev_position = position
```

The first frame counts as 1, and each later frame delta covers all of the skipped frames too (no special cases needed). This also avoids reading `self._start_pos` which is set on the decode thread.

Breakthrough:
Hi @nathanmdan; just checking in, are you okay making the changes I asked for?

nathanmdan:
Hi @Breakthrough, my apologies, but can you please show me where I might find your change requests? I'm afraid I don't see them.

Breakthrough (changes requested):

Breakthrough:
Sorry about that - I thought they had already been published. My apologies for the mixup. You should see them now.

nathanmdan:
No problem, I'll look into this!

scenedetect/scene_manager.py
nathanmdan: This revision looks good to me. Thank you for the the suggestion!

scenedetect/scene_manager.py
```diff
@@ -587,7 +588,11 @@ def detect_scenes(
                         progress_bar.set_description(
                             PROGRESS_BAR_DESCRIPTION % len(self._cutting_list), refresh=False
                         )
-                    progress_bar.update(1 + frame_skip)
+                    # Increment progress bar by delta of position.frame_num instead of 1
+                    # to handle VFR video where frame count is an approximation.
+                    delta = 1 if prev_position is None else position.frame_num - prev_position.frame_num
```
Breakthrough: Thanks @nathanmdan, this is exactly it - the sentinel covers the seek case, and the deltas absorb frame_skip on their own.

One thing before I merge: that ternary is 104 characters, and CI runs `ruff format --check` at a 100 limit, so Static Analysis will fail. `ruff format` wraps it as:

```suggestion
                    delta = (
                        1 if prev_position is None else position.frame_num - prev_position.frame_num
                    )
```

You can catch these locally with:

    pip install -e .[dev]
    ruff check scenedetect/ tests/
    ruff format scenedetect/ tests/

Also, please ignore the failing Actions runs on your fork - those are deploy workflows (docs, website, Docker) trying to push to branches and registries that only exist on the upstream repo, so they can never pass on a fork. I guarded them in fdd077a; they will go quiet once main is merged in again.  You can also resolve the open comments now, thanks.

Breakthrough (approved):

Breakthrough:
Thank you for the pull request!

nathanmdan:
Of course! Thank you for the suggestions.

PR #563, opened 2026-08-08
Add deprecation warnings for legacy framerate aliases

### LINKED ISSUE #548: Add DeprecationWarnings for deprecated APIs in v0.7 release
To avoid too much churn with the v0.7 release, some breaking changes allow fallback to old property names (e.g. passing `framerate` instead of the new `frame_rate`).  We should upgrade these to runtime warnings in v0.8 so they can be fully removed in v0.9.

MuhammadBilalKhan267:
Closes #548

## Summary

* Emit `DeprecationWarning` for deprecated `framerate` aliases across `FrameTimecode`, `open_video()`, video backends, and the CLI.
* Preserve existing compatibility behavior, including `frame_rate` taking precedence when both forms are provided.
* Add/update tests for the deprecated aliases, including `VideoCaptureAdapter`.
* Update `get_framerate()` to direct users to `frame_rate` without emitting duplicate warnings.

## Testing

* Added warning assertions for deprecated aliases.
* Verified canonical `frame_rate` usage remains warning-free.
* Verified legacy aliases continue to work while emitting `DeprecationWarning`.

### DIFF
--- a/docs/cli.rst
+++ b/docs/cli.rst
@@ -67,7 +67,7 @@ Options
 
 .. option:: --framerate FPS
 
-  [DEPRECATED] Use :option:`-f/--frame-rate <-f>` instead.
+  Alias of :option:`-f/--frame-rate <-f>`.
 
 .. option:: -m TIMECODE, --min-scene-len TIMECODE
 
--- a/scenedetect/_cli/__init__.py
+++ b/scenedetect/_cli/__init__.py
@@ -22,6 +22,7 @@ import inspect
 import logging
 import os
 import os.path
+import warnings
 from copy import copy
 
 import click
@@ -361,8 +362,13 @@ def scenedetect(
     ctx = ctx.obj
     assert isinstance(ctx, CliContext)
 
-    # TODO(https://scenedetect.com/issue/548): emit DeprecationWarning when `--framerate`
-    # is used, once downstream users have had a release to migrate to `--frame-rate`.
+    if framerate_legacy is not None:
+        warnings.warn(
+            "`--framerate` is deprecated; use `--frame-rate` instead.",
+            DeprecationWarning,
+            stacklevel=2,
+        )
+
     if frame_rate is None:
         frame_rate = framerate_legacy
     elif framerate_legacy is not None:
--- a/scenedetect/_cli/__init__.py
+++ b/scenedetect/_cli/__init__.py
@@ -22,7 +22,6 @@ import inspect
 import logging
 import os
 import os.path
-import warnings
 from copy import copy
 
 import click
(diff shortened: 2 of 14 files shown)


scenedetect/_cli/__init__.py
```diff
@@ -361,8 +362,13 @@ def scenedetect(
     ctx = ctx.obj
     assert isinstance(ctx, CliContext)
 
-    # TODO(https://scenedetect.com/issue/548): emit DeprecationWarning when `--framerate`
-    # is used, once downstream users have had a release to migrate to `--frame-rate`.
```
Breakthrough, line 365: I know I originally added this comment, but can we explore what it would take to just keep both working and make them aliases of one another?  (e.g. have `-f` `--framerate` and `--frame-rate` all be aliases of one another).  I think click supports this.

Breakthrough (changes requested):
For the CLI, the tricky part is I'd like to keep both as hidden aliases of one another.  This might involve some trickery with the config parser unfortunately, but if you can find a better way feel free. Let me know if you have any questions.

API changes LGTM, thanks for the pull request!

MuhammadBilalKhan267:
I updated the hidden `--framerate` option to map directly to `frame_rate` and removed the separate `framerate_legacy` parameter. All three forms (`-f`, `--frame-rate`, and `--framerate`) now act as aliases. If multiple forms are provided, Click uses the last value, which seems appropriate since `--framerate` remains supported rather than deprecated. For example, `--frame-rate 30 --framerate 24` resolves to 24, whereas previously `--frame-rate` always took precedence.

docs/cli.rst
```diff
@@ -67,7 +67,7 @@ Options
 
 .. option:: --framerate FPS
 
-  [DEPRECATED] Use :option:`-f/--frame-rate <-f>` instead.
+  Alias of :option:`-f/--frame-rate <-f>`.
```
Breakthrough, line 70: Nit: We should add a TODO somewhere (or comment on the bug) to merge this help text into the prev. one or omit it eventually.

scenedetect/_cli/__init__.py
```diff
@@ -237,14 +237,15 @@ def print_command_help(ctx: click.Context, command: click.Command):
     default=None,
     help="Override frame rate with value as frames/sec.",
 )
+# Keep --framerate separate so Click can hide it while mapping both spellings to frame_rate.
 @click.option(
     "--framerate",
-    "framerate_legacy",
+    "frame_rate",
```
Breakthrough, line 243: I hadn't realized that we could use the same variable, does this cause any issues?  It seems like this should work as intended though.

Breakthrough (approved):

scenedetect/_cli/__init__.py
MuhammadBilalKhan267: It simply uses the last value provided. Only behavior change is that `--frame-rate 30 --framerate 24` now resolves to 24 instead of 30.

docs/cli.rst
MuhammadBilalKhan267: Wouldn't merging it with `--frame-rate` increase visibility? For a hidden alias, I think we should remove it. We can also treat `--frame-rate` and `--framerate` as equal aliases. This requires only one Click option as it supports aliases.

Breakthrough: Good point. Let's do the latter if possible (equal aliases).  We can merge this PR as-is and tackle it as a follow up.

Breakthrough:
Approved, thanks for the PR!

Issue #534, opened 2026-02-10
Export scene change to ``# keyframe format v1``


moi15moi:
**Problem/Use Case**

Export the generated scene list to a file and import them to [aegisub](https://github.com/TypesettingTools/Aegisub).

**Solutions**

Be able to export ``# keyframe format v1`` file with the cli.
This is a old format that some tool, like aegisub, use to represent the scene detection.

**Examples:**

``keyframe format v1`` format is really simple.
The first 2 lines are always theses (you can technically set an fps, but it doesn't do anything. Most parser just skip the ``fps`` line):
```
# keyframe format v1
fps 0
```
After that, each line will contain the frame number where a scene change is detected.
Ex:
```
# keyframe format v1
fps 0
0
120
156
343
```

Breakthrough:
Does this format work with VFR video / is it presentation time or actual frame number?  Can certainly add this in a future version.

moi15moi:
It is frame number, so yes, it work really well with vfr.

MuhammadBilalKhan267:
I would like to work on this.

MuhammadBilalKhan267:
I found one issue with VFR input while implementing this.

Detection processes frames sequentially, but cuts retain timestamps rather than decoded frame ordinals. `FrameTimecode.frame_num` converts the timestamp using the average frame rate, which can produce incorrect frame numbers for VFR video.

There are two possible approaches:

1. Make a second pass with `decode=False` and map each cut timestamp to its actual zero-based frame ordinal. This is isolated, but requires reading the video twice and assumes the input can be replayed.
2. Retain exact ordinals during the initial decode. This requires broader pipeline changes because buffered or calculated cuts must preserve their ordinals. Starting detection partway through a video with `time --start` also makes the absolute ordinal harder to determine.

I can handle this in a follow-up. If either approach seems like a better fit for the project, feel free to let me know.

PR #565, opened 2026-08-12
Add keyframe format v1 CLI export

MuhammadBilalKhan267:
## Summary

Adds a `save-keyframes` CLI command that exports detected scene changes using keyframe format v1.

The command supports custom filenames and output directories through CLI options and `scenedetect.cfg`.

## Testing

Added CLI tests covering:

- Default output
- Custom filenames
- Detection with a start offset

### Known limitation

Frame numbers are currently approximate for VFR input due to the existing `FrameTimecode.frame_num` behavior. 
Exact VFR frame ordinal support is tracked in #566.

Addresses #534


Breakthrough:
> VFR inputs may produce incorrect frame numbers because cuts currently store timestamps and FrameTimecode.frame_num derives a frame number using the average frame rate. Exact VFR frame ordinal support will be handled separately.

Can we add a test case somehow for downstream usage of the file format to ensure it's valid?  E.g. if ffmpeg accepts it can we run it to verify the output.  I would like VFR handled in this PR if possible.

MuhammadBilalKhan267:
I think the VFR problem deserves it's own issue. `frame_num` gets used in `save_qp` and `threshold_detector` as well.

Breakthrough (approved):

Breakthrough:
Thanks for the PR!

PR #566, opened 2026-08-14
Accept a path or file handle in write_scene_list

jennifferweslowski-design:
Fixes #523

`write_scene_list("scenes.csv", scenes)` currently raises `TypeError: argument 1 must have a "write" method` because the path string is passed straight to `csv.writer`.

This change lets the first argument be a file handle (existing callers) or a `str` / `pathlib.Path`. A path is opened with `newline=""` inside a context manager so the file is closed after writing, which matches the concern on the issue.

Tests cover a `StringIO` handle, a `str` path, and a `Path`. No video fixtures.

I could not run the suite in this environment; the new tests follow the existing `_fake_scenes` pattern in `tests/test_output.py`.

PR #570, opened 2026-09-01
[output] Support paths and file handles in scene-list writers

### LINKED ISSUE #567: Use Decorator Pattern for Consistent Output Handling
Right now [the naming *and* type convention for output functions' arguments](https://www.scenedetect.com/docs/latest/api/output.html) is somewhat inconsistent.  As identified in #523, there exist other common patterns and precedents for taking paths as well as file handles. Some functions also take only `Path`, and others take only `TextIO`.

It might be worth investigating if we can create a decorator to annotate functions to avoid these issues. This would allow the implementation itself to accept `file` as a file handle, and not have to worry about path processing. I'm personally a fan of [numpy's simple definition as just `file:
file, str, or pathlib.Path`](https://numpy.org/doc/stable/reference/generated/numpy.save.html#numpy.save) for the public API, and have the decorator "convert" that into a file handle for the internal implementation (e.g. handling the context).

We should consider something similar if a function requires a directory or other kinds of templates *if* it makes sense.

MuhammadBilalKhan267:
Closes #567.

- Add a shared decorator for opening and closing output paths.
- Support `str`, `Path`, and open text files for EDL, FCPXML, FCP7 XML, and OTIO output.
- Preserve CSV newline handling.
- Add tests and update API documentation and changelog.

Tests: 382 passed, 2 skipped.

### DIFF
--- a/scenedetect/output/__init__.py
+++ b/scenedetect/output/__init__.py
@@ -17,9 +17,12 @@ CSV/HTML, or splitting the input video into individual shots.
 
 import csv
 import functools
+import inspect
+import io
 import json
 import logging
 import math
+import os
 import typing as ty
 from fractions import Fraction
 from pathlib import Path
@@ -65,6 +68,7 @@ from scenedetect.output.video import (
 from scenedetect.output.video import (
     split_video_mkvmerge as split_video_mkvmerge,
 )
+from scenedetect.platform import BytePath, StrPath
 
 logger = logging.getLogger("pyscenedetect")
 
@@ -72,36 +76,59 @@ logger = logging.getLogger("pyscenedetect")
 _F = ty.TypeVar("_F", bound=ty.Callable[..., ty.Any])
 
 
-def _open_output_file(parameter_name: str, *, newline: str | None = None):
-    """Allows a writer's first argument to be an open text file or an output path.
+def _open_output_file(output_format: str) -> ty.Callable[[_F], _F]:
+    """Allows a writer's first argument to be an open text file or a filesystem path.
 
-    File handles remain open after the writer returns. Paths are opened in write mode and closed
-    when the writer returns.
+    File handles remain open after the writer returns. Path-based output is rendered in memory
+    before the destination is opened, then written as UTF-8 with no newline translation.
     """
 
     def decorator(func: _F) -> _F:
+        parameters = tuple(inspect.signature(func).parameters.values())
+        assert parameters and parameters[0].kind in (
+            inspect.Parameter.POSITIONAL_ONLY,
+            inspect.Parameter.POSITIONAL_OR_KEYWORD,
+        )
+        parameter_name = parameters[0].name
+
         @functools.wraps(func)
-        def wrapper(*args: ty.Any, **kwargs: ty.Any):
+        def wrapper(*args: ty.Any, **kwargs: ty.Any) -> ty.Any:
             output_file = args[0] if args else kwargs[parameter_name]
 
-            if not isinstance(output_file, (str, Path)):
+            if not isinstance(output_file, (str, bytes, os.PathLike)):
+                if not callable(getattr(output_file, "write", None)):
+                    raise TypeError(
+                        f"{parameter_name} must be a filesystem path or writable text file"
+                    )
+                logger.info(
+                    "Writing scenes in %s format to %s",
+                    output_format,
+                    getattr(output_file, "name", "<file-like object>"),
+                )
                 return func(*args, **kwargs)
 
-            with open(output_file, "w", newline=newline) as file_handle:
+            with io.StringIO(newline="") as output_buffer:
                 if args:
-                    return func(file_handle, *args[1:], **kwargs)
+                    result = func(output_buffer, *args[1:], **kwargs)
+                else:
+                    kwargs[parameter_name] = output_buffer
+                    result = func(**kwargs)
+                contents = output_buffer.getvalue()
 
-                kwargs[parameter_name] = file_handle
-                return func(**kwargs)
+            logger.info("Writing scenes in %s format to %s", output_format, output_file)
+            with open(output_file, "w", encoding="utf-8", newline="") as file_handle:
+                file_handle.write(contents)
+
+            return result
 
         return ty.cast(_F, wrapper)
 
     return decorator
 
 
-@_open_output_file(parameter_name="output_csv_file", newline="")
+@_open_output_file("CSV")
 def write_scene_list(
-    output_csv_file: str | Path | ty.TextIO,
+    output_csv_file: StrPath | BytePath | ty.TextIO,
     scene_list: SceneList,
     include_cut_list: bool = True,
     cut_list: CutList | None = None,
@@ -125,8 +152,8 @@ def write_scene_list(
     Raises:
         TypeError: "delimiter" must be a 1-character string
     """
-    output_csv_file = ty.cast(ty.TextIO, output_csv_file)
-    csv_writer = csv.writer(output_csv_file, delimiter=col_separator, lineterminator=row_separator)
+    output_file = ty.cast(ty.TextIO, output_csv_file)
+    csv_writer = csv.writer(output_file, delimiter=col_separator, lineterminator=row_separator)
     # If required, output the cutting list as the first row (i.e. before the header row).
     if include_cut_list:
         csv_writer.writerow(
@@ -326,9 +353,9 @@ def _parse_edl_start_timecode(value: str, frame_rate: Fraction | float) -> int:
     return round((hours * 3600 + minutes * 60 + seconds) * float(frame_rate)) + frames
 
 
-@_open_output_file("output_path")
+@_open_output_file("EDL")
 def write_scene_list_edl(
-    output_path: str | Path | ty.TextIO,
+    output_path: StrPath | BytePath | ty.TextIO,
     scene_list: SceneList,
     title: str = "PySceneDetect",
     reel: str = "AX",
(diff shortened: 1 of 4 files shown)


scenedetect/output/__init__.py
```diff
@@ -68,8 +69,39 @@
...
+
+    def decorator(func: _F) -> _F:
+        @functools.wraps(func)
+        def wrapper(*args: ty.Any, **kwargs: ty.Any):
+            output_file = args[0] if args else kwargs[parameter_name]
+
+            if not isinstance(output_file, (str, Path)):
+                return func(*args, **kwargs)
+
+            with open(output_file, "w", newline=newline) as file_handle:
```
Breakthrough: This will open and truncate the file before the writer operates, which means any exceptions during generation leave an empty file or clobber an existing one.

Can we render the contents of the file in memory first, only opening + writing it to disk once the body is constructed? Most writers build the whole document in memory so this should be relatively cheap.

scenedetect/output/__init__.py
```diff
@@ -68,8 +69,39 @@
...
+    File handles remain open after the writer returns. Paths are opened in write mode and closed
+    when the writer returns.
+    """
+
+    def decorator(func: _F) -> _F:
+        @functools.wraps(func)
+        def wrapper(*args: ty.Any, **kwargs: ty.Any):
+            output_file = args[0] if args else kwargs[parameter_name]
+
+            if not isinstance(output_file, (str, Path)):
```
Breakthrough: This doesn't necessarily work if the OS uses bytes for paths. Previously we coerced the input into `Path(output_path)` to ensure the type - consider using `isinstance(output_file, (str, bytes, os.PathLike))` here and annotating the writers as `StrPath | ty.TextIO`.

scenedetect/output/__init__.py
Breakthrough: I think we need to add an explicit encoding here, UTF-8 seems like a good candidate (set `encoding="utf-8"` here).

scenedetect/output/__init__.py
```diff
@@ -68,8 +69,39 @@
...
+
+                kwargs[parameter_name] = file_handle
+                return func(**kwargs)
+
+        return ty.cast(_F, wrapper)
+
+    return decorator
+
+
+@_open_output_file(parameter_name="output_csv_file", newline="")
```
Breakthrough: The CSV writer passes `newline=""` but the other four open with the default value, so on Windows the same writer could emit CRLF when given a path, but then LF when given a handle.

Path-mode CRLF was pre-existing, but we should make the behaviour consistent. Since every writer emits its own line endings, maybe we could just open with `newline=""` unconditionally in the decorator and drop `newline` for this special case. Thoughts?

scenedetect/output/__init__.py
```diff
@@ -68,8 +69,39 @@
...
+    """Allows a writer's first argument to be an open text file or an output path.
+
+    File handles remain open after the writer returns. Paths are opened in write mode and closed
+    when the writer returns.
+    """
+
+    def decorator(func: _F) -> _F:
+        @functools.wraps(func)
+        def wrapper(*args: ty.Any, **kwargs: ty.Any):
+            output_file = args[0] if args else kwargs[parameter_name]
```
Breakthrough, line 96: Can we use a single checked source of truth for the filename somehow?  Or just drop the parameter name entirely and assert at decoration time that the first param is the file always?

scenedetect/output/__init__.py
```diff
@@ -93,17 +125,7 @@ def write_scene_list(
...
-            write_scene_list(
-                file_handle,
-                scene_list,
-                include_cut_list=include_cut_list,
-                cut_list=cut_list,
-                col_separator=col_separator,
-                row_separator=row_separator,
-            )
-        return
+    output_csv_file = ty.cast(ty.TextIO, output_csv_file)
```
Breakthrough: Can we just assert the instance is of type X instead of casting here and elsewhere? (e.g .`assert not isinstance(output_path, (str, os.PathLike))`)

This avoids type errors when someone passes in the wrong type.

Also: some writers reassign the arg in place, while the other four introduce a new `output_file` name. Might be worth consistency here to guide a future API change, since I do think the current arg name is too verbose. That's for another time though of course.

scenedetect/output/__init__.py
```diff
@@ -333,14 +357,16 @@ def write_scene_list_edl(
         in_tc = _edl_timecode(start + offset_frames)
         out_tc = _edl_timecode(end + offset_frames)
         lines.append(f"{(i + 1):03d}  {reel} V     C        {in_tc} {out_tc} {in_tc} {out_tc}")
-    logger.info("Writing scenes in EDL format to %s", output_path)
-    with open(output_path, "w") as f:
-        # `scenedetect` is imported lazily to avoid a circular import at module load.
-        import scenedetect
+    logger.info(
+        "Writing scenes in EDL format to %s",
```
Breakthrough: We do the getattr thing a lot here, can we just add this to the decorator to avoid repeating it?

scenedetect/output/__init__.py
```diff
@@ -68,8 +69,39 @@
 logger = logging.getLogger("pyscenedetect")
 
 
+_F = ty.TypeVar("_F", bound=ty.Callable[..., ty.Any])
+
+
+def _open_output_file(parameter_name: str, *, newline: str | None = None):
```
Breakthrough: Nit: consider adding a return type annotation here and for the inner `wrapper` for consistency with the other definitions in this file.

tests/test_output.py
```diff
@@ -291,12 +291,31 @@ def test_write_scene_list_edl(tmp_path: Path):
...
-    """`output_path` must accept both Path and str."""
+    """EDL output accepts a string path."""
     scenes = _fake_scenes(_FPS_CFR, [(0, 30)])
     output_path = tmp_path / "scenes.edl"
     write_scene_list_edl(str(output_path), scenes)
     assert output_path.exists()
+    assert "TITLE:" in output_path.read_text()
+
+
+def test_write_scene_list_edl_accepts_path(tmp_path: Path):
```
Breakthrough: The four new *_accepts_path tests seem to be subsets of existing tests directly above them. The remaining eight could also be parameterized to avoid so much code duplication. Consider deleting the `_accepts_path` tests and instead parameterize over the target kind (file handle, string, os.Path, etc). See other uses of `@pytest.mark.parametrize` in this file for examples.

Breakthrough (changes requested):
Thanks for getting started with this one, there's a lot of moving parts here.

scenedetect/output/__init__.py
MuhammadBilalKhan267: I had noticed this earlier but followed the pattern introduced in #566, where a path is opened in write mode before delegating to the writer. Changed now.

scenedetect/output/__init__.py
MuhammadBilalKhan267: Done. Now we get the name of the file parameter from the function signature.

scenedetect/output/__init__.py
MuhammadBilalKhan267: The decorator converts paths to StringIO before calling the writer and rejects objects without write(). The suggested assertion would therefore always pass. ty.cast() only tells Pyright that this is a text file.

tests/test_output.py
MuhammadBilalKhan267: I initially followed the separate file-handle/string/Path test pattern from #566. Changed now.

scenedetect/output/__init__.py
Breakthrough: Assertions about meta attributes should also fulfill Pyright's resolution. The reason I prefer the assertion is that it ensures we fail loudly if for whatever reason someone tries to change it incorrectly.

MuhammadBilalKhan267: I still think the decorator validation plus `ty.cast()` is clearer, but I’ve switched to the defensive assertion as requested.

scenedetect/output/__init__.py
```diff
@@ -93,18 +152,9 @@ def write_scene_list(
...
-                file_handle,
-                scene_list,
-                include_cut_list=include_cut_list,
-                cut_list=cut_list,
-                col_separator=col_separator,
-                row_separator=row_separator,
-            )
-        return
-    csv_writer = csv.writer(output_csv_file, delimiter=col_separator, lineterminator=row_separator)
+    assert not isinstance(output_csv_file, (str, bytes, os.PathLike))
```
Breakthrough, line 155: Is there a positive type assertion we can use here and elsewhere?  I think the idiomatic way to do this is:

```python
assert isinstance(output_csv_file, io.TextIOBase)
```

Please also add a clarifying comment (can be inline, or part of the assertion) of something like "the _open_output_file decorator transforms the type for us").

Breakthrough (approved):
Approving and merging as-is but please see the last comment I added for follow up.

Thanks for the PR!

scenedetect/output/__init__.py
```diff
@@ -64,12 +68,67 @@
...
 )
+from scenedetect.platform import BytePath, StrPath
 
 logger = logging.getLogger("pyscenedetect")
 
 
+_F = ty.TypeVar("_F", bound=ty.Callable[..., ty.Any])
+
+
+def _open_output_file(output_format: str) -> ty.Callable[[_F], _F]:
```
Breakthrough, line 79: Could we also add a unit test that ensures we don't create a file in the case the output generator throws an exception?  It might be worth having some tests specifically for this decorator, and would make it more obvious what it's doing under the hood (since then the constraints + usage would have simpler/terse examples).

MuhammadBilalKhan267: yes sure

scenedetect/output/__init__.py
MuhammadBilalKhan267: The positive assertion would reject custom file-like objects that implement write but don't inherit from TextIOBase.

PR #571, opened 2026-09-11
[bugfix] Apply AdaptiveDetector min_scene_len to the emitted cut (#408)

### LINKED ISSUE #408: `min_scene_len` parameter isn't respected with AdaptiveDetector
Hello. First of all, thanks for all the work you've done in this project. I've been using PyScene for some months until now and it has been very helpful, intuitive and easy to use.

**Description of the issue**
I'm having trouble with the parameter `min_scene_len` because it doesn't seem to do nothing if I change it.

I'm using the AdaptiveDetector module to differenciate shots in videos and one problem I get a lot is that when some frame surpases the threshold, normally there are some following frames also surpasing the threshold until it decays, and these additional frames trigger the scene detection. So I end up with 2 to 3 very short scenes whenever there is a scene change.

My approach to handle this was to use the `min_scene_len` parameter to filter this short scenes but, even with this parameter, the module generates clips with fewer frames than the specified parameter.

The code I'm running to get the scenes is this one:
```py
scenemanager = SceneManager(stats_manager=StatsManager())
scenemanager.add_detector((AdaptiveDetector(adaptive_threshold=2, window_width=20, min_scene_len=24)))
scenemanager.detect_scenes(open_video(video))
scenes = scenemanager.get_scene_list()
```
Then I cut the scenes with this ffmpeg command:
```py
for idx, scene in enumerate(scenes):
        output_name = os.path.join(output_basename, f"scene_{idx}.mp4")
        periods[idx] = {}
        start_time = scene[0].get_seconds()
        end_time = scene[1].get_seconds()
        subprocess.run(f"ffmpeg -y -i {video_name} -ss {start_time} -to {end_time} -c:v libx264 -preset medium -crf 30 -c:a copy {output_name}", shell=True)
```
Which writes all the clips in my project folder.

Some of the clips (especially the ones I told in shots boundaries) are very short (0.1 to 0.4 seconds) and it didn't seem to be coherent with the `min_scene_len=24` parameter.

I checked the number of frames of these videos with ffprobe, obtaining this result:
```sh
ffprobe -v error -select_streams v:0 -count_frames -show_entries stream=nb_read_frames -of default=nokey=1:noprint_wrappers=1 video_clip.mp4
4
```
So my suspicions were confirmed. I'm having clips shorter than the specified minimum scene length.

**Question**

Is there something am I doing the wrong way or maybe I don't understand correctly the use of this parameter? Or is this a bug I found? I didn't manage to find anything about this topic, especially with the AdaptiveDetector module.

Thank you!

cpruijsen:
## Summary

- `AdaptiveDetector` was checking `min_scene_len` against the *current* frame while emitting the cut `window_width` frames earlier. With the reporter's `window_width=20` and `min_scene_len=24`, a second peak could be written 4 frames after the previous cut (issue #408).
- The length check now uses the emitted target frame, which is what 0.6.4 (`d8397bc`) intended but did not actually change (the comparison was left on the current frame after a rename).

Provenance: the owner identified the current-vs-target mismatch on the issue; collaborator @wjs018 described the 100/124/104 example; the reporter confirmed that example and measured a 4-frame clip with ffprobe.

## Decision

- **Chose:** Compare `min_scene_len` to `target_timecode` (the cut that is actually returned).
- **Alternative:** Keep the current-frame comparison and add `window_width` to the threshold, or move AdaptiveDetector onto `FlashFilter` like `ContentDetector`.
- **Why:** Smallest change that matches the owner's diagnosis and the stated intent of `d8397bc`. Happy to switch to `FlashFilter` if you want merge-mode behaviour here.

A second judgement call: the regression test uses synthetic luma steps with the reporter's `window_width=20` / `min_scene_len=24`, not `goldeneye.mp4`. Well-spaced cuts on that clip never enter the `min_scene_len - window_width` window, so a goldeneye-only assertion would pass with or without the fix.

## Checklist

- [x] New AdaptiveDetector unit test fails without the fix (`[100, 104, 140, 144]` vs `[100, 140]`) and passes with it
- [x] `ruff check` / `ruff format` on the touched Python files
- [x] Follows the Google Python Style Guide
- [x] Changelog entry under 0.7.2 (development section)

## Test plan

- [x] Synthetic luma cuts at frames 100/104 and 140/144 with `window_width=20`, `min_scene_len=24`: without the fix the detector emits `[100, 104, 140, 144]`; with the fix it emits `[100, 140]`.
- [x] `pytest tests/test_detectors.py::test_detectors_with_stats` (all detectors including AdaptiveDetector, with a StatsManager)
- [ ] Optional: `detect-adaptive` on a clip with a decaying double peak and `--min-scene-len 24 --frame-window 20`; no output scene should be shorter than 24 frames.

Fixes #408


Breakthrough (approved):

Breakthrough:
Thanks for the PR!

PR #572, opened 2026-09-12
[output] Add output decorator regression coverage

MuhammadBilalKhan267:
Follow-up to #570.

Adds focused coverage for `_open_output_file` to ensure failed output generation does not create the destination file.

Also documents that the decorator replaces filesystem paths with writable text streams before invoking each writer.

Tests: 390 passed, 2 skipped


Breakthrough (approved):
Thanks for fixing this up!

Breakthrough (approved):

PR #574, opened 2026-09-14
Add a minimum fade-out duration to ThresholdDetector

### LINKED ISSUE #278: Add option to `detect-threshold` to ignore shorter scene boundaries
**Description of Problem & Solution**

Sometimes there are shorter fades-through-black in the middle of a scene and you don't want to introduce cut points at them. Currently the `process_frame` method of `ThresholdDetector` will introduce a scene cut any time the frame average dips below the threshold.

**Proposed Implementation:**

Add an optional argument `--min-cut-len=[time]` to `detect-threshold` that will skip (that is, not cut at) fades-through-black that are shorter than *time*. If *time* is an integer, interpret it as a number of frames; if *time* is a floating-point number with `s` appended, interpret it as a number of seconds; if *time* is a time code, interpret it as a hour:minute:second duration.

I think all that would have to be done is to set a property `min_cut_len` and change line [line 149 of `scenedetect.detectors.threshold_detector`](https://github.com/Breakthrough/PySceneDetect/blob/master/scenedetect/detectors/threshold_detector.py#L149) from

            elif self.last_fade['type'] == 'out' and frame_avg >= self.threshold:

to something along the lines of

            elif self.last_fade['type'] == 'out' and (frame_num - self.last_fade['frame'] > self.min_cut_len) and frame_avg >= self.threshold:

**Alternative Solutions:**

Setting the minimum *scene* length with `--min-scene-len` or `-m` will sometimes have the desired effect, but only coincidentally. If a 20-minute scene has a half-second fade five minutes from the beginning, setting `-m=360s` will not cut at that point. But setting this minimum scene length too high may miss the longer fades and effectively clump different parts of different scenes together.

aprylewu:
Add an optional `min_out_length` to `ThresholdDetector`, exposed as `detect-threshold --min-out-length` and the matching configuration setting. This implements the duration filter discussed in #278 using the name suggested there.

Fixes #278.

For example, with a four-frame minimum, a two-frame dark interval is ignored while a later six-frame interval produces its normal fade-biased cut. Ignoring the first interval still updates the detector's fade state, so the later cut uses its own fade-out position. The same filter applies to a final fade-out, counting the last processed frame; a trailing four-frame interval meets a four-frame minimum.

The default is zero and preserves existing behavior. The new constructor argument follows the existing arguments to preserve positional calls. It accepts the same duration representations as `min_scene_len` and leaves that setting's existing cut-spacing behavior unchanged.

Validation:
- New regression tests cover threshold boundaries, successive short/long fades, both threshold methods, fade bias, minimum scene length, final fades, CLI/config parsing, and real synthetic videos decoded with OpenCV and PyAV.
- The complete standard test suite passed: 426 passed, 2 skipped, 83 release tests deselected. The new test module contributes 37 passing cases.
- `ruff check .` and `ruff format --check .` passed for all 104 Python files.
- An independent comparison against the base detector found identical default results and state across 1,920 configurations.

The CLI reference was regenerated with `docs/generate_cli_docs.py`. This change uses the existing nominal-frame endpoint convention for the last frame; it does not change general VFR endpoint handling.

Developed with OpenAI Codex assistance and independently reviewed by another coding agent. All reported checks were run locally.


tests/test_threshold_detector.py
Breakthrough, line 1: File headers must all use standardized format:
```
#
#            PySceneDetect: Python-Based Video Scene Detector
#   -------------------------------------------------------------------
#     [  Site:    https://scenedetect.com                           ]
#     [  Docs:    https://scenedetect.com/docs/                     ]
#     [  Github:  https://github.com/Breakthrough/PySceneDetect/    ]
#
# Copyright (C) 2026 Brandon Castellano <http://www.bcastell.com>.
# PySceneDetect is licensed under the BSD 3-Clause License; see the
# included LICENSE file, or visit one of the above pages for details.
#
```

Breakthrough (changes requested):
Thanks for the PR.  Should we set a reasonable default value for this as well in the CLI?  How often does this occur?

tests/test_threshold_detector.py
aprylewu: Updated to the exact header format you provided in 8c1d7df. Also replaced the OpenCV fourcc alias in the test with `cv2.VideoWriter.fourcc`, which fixes the Pyright failure without changing the generated video.

aprylewu:
Thanks for reviewing. I don't have representative real-video measurements of how often this occurs. The motivation is the brief within-scene fades described in #278; my validation uses synthetic clips, so it establishes the filter behavior rather than its prevalence.

A nonzero CLI default is possible, but the duration measured here is time *below the brightness threshold*, not the whole fade. Even a deliberate fade-through-black can have a very short below-threshold interval. That's why I initially kept the default at zero.

If we want suppression enabled by default in the CLI, `0.1s` seems like a conservative starting point to evaluate, with the API default still at zero for compatibility. That would be a proposed heuristic, not a value established by the tests, and it would still need representative clips to assess missed transitions. My preference is to keep zero for now unless there is an existing sample set we can use to choose the default.

The header and CI type-check issue are fixed in 8c1d7df. Local Pyright now reports zero errors; Ruff checks pass. The standard test suite also passed again: 426 passed, 2 skipped, 83 release tests deselected.

aprylewu:
Went ahead and set the CLI default for `min-out-length` to `0.1s` in fe725f2, while leaving the `ThresholdDetector` API default at `0` for compatibility. Docs and the config example are updated to match.

`pytest -q tests/test_threshold_detector.py`: 37 passed.

Breakthrough (approved):
Thanks for the PR!

aprylewu:
Thanks for the review and for merging!

